PR #7, opened 2020-05-13
134 rake task max

maxkadel:
If you go into the running web container and run the rake command, it will populate your Solr core with the existing fixture documents (separate PR).

Future development includes adding dotenv and refining how we define the Solr url and cores (service name, e.g. "solr" or "db" vs. localhost)

edit: Not passing CI. using "localhost" for Solr in CI results in `RSolr::Error::Http - 404 Not Found`, using "solr" for Solr in CI results in `RSolr::Error::ConnectionRefused`. But it works locally!

image


martinlovell (approved):
Went through with Max, looks good.

PR #14, opened 2020-05-21
try pushing using circle ci

maxkadel:


maxkadel:
close - going with workflows for now

PR #26, opened 2020-05-27
Add private/restricted fixtures plus specs

dylansalay:
- Created 2 mock private fixtures and 2 mock restricted 'yale-only' fixtures
- Added specs to assess the private and restricted fixtures


spec/services/fixture_indexing_service_spec.rb
```diff
@@ -27,10 +26,41 @@
...
 
-  it "can index a fake private file to Solr" do
-    FixtureIndexingService.index_to_solr(priv_oid)
-    solr = RSolr.connect url: "#{solr_url}/#{solr_core}"
-    response = solr.get 'select', params: { q: '*:*' }
-    expect(response["response"]["docs"].first.dig("id")).to eq("16189097-priv")
+  context "it can index private files" do
+    let(:priv_oid) { "16189097-priv" }
+
+    it "can determine if a file is private" do
```
maxkadel: I don't think this is actually determining that the file is private, is it? It's just mapping the defined ID to the defined object.

spec/services/fixture_indexing_service_spec.rb
```diff
@@ -27,10 +26,41 @@
...
+      solr = RSolr.connect url: "#{solr_url}/#{solr_core}"
+      response = solr.get 'select', params: { q: '[private copy]' }
+      expect(response["response"]["numFound"]).to eq 2
+    end
+  end
+
+  context "it can index yale-only files" do
+    let(:rest_oid) { "2107188-yale" }
+
+    it "can determine if a file is restricted to Yale only" do
```
maxkadel: Same issue as the private, the test is not doing what it says.

maxkadel (changes requested):
I would like the fixture records to have a separate field that is indexed to Solr that is only for the visibility of the object. Let's use `visibility_ssi`.

I'd like to see a test that adds multiple versions of the same object (so the "public" and "yale-only" versions of "Fair Lucretia’s garland", and then searches based on the title, and which then can distinguish between the public and private copies of that object based on something other than the title or ID). 

I believe you may need to change the oids in the yale-only and private fixture files in order to do that.

maxkadel (approved):

bess (approved):

PR #28, opened 2020-05-28
use the same Solr mapping as Blacklight, same ordering

maxkadel:
* Use latest releases of other applications
* Use the same Solr mapping as Blacklight
* Re-order fields so they're easy to compare to Blacklight catalog controller
* Those that don't yet have a clear mapping from the Metadata Cloud are commented out


.env
```diff
@@ -12,12 +12,12 @@ POSTGRES_PASSWORD=password
 
 # Docker Image tags
 MANAGEMENT_TAG=beta001
-MANAGEMENT_BASE_TAG=latest
-BLACKLIGHT_BASE_TAG=105ac
```
mark-dce, line 16: It doesn't seem like the management app should need to be concerned with the BLACKLIGHT_BASE_TAG.  But that's a topic for  a different PR.

mark-dce (approved):
👍

PR #38, opened 2020-06-03
162 solr index button

dylansalay:
- Button successfully indexes metadata fixtures

Screen Shot 2020-06-03 at 12 15 30 PM


app/views/solr_task/run_task.html.erb
```diff
@@ -0,0 +1 @@
+<h1> Your files have been indexed </h1>
```
maxkadel, line 1: Would you make this a flash message, rather than its own page?

app/controllers/solr_task_controller.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class SolrTaskController < ApplicationController
+  def index; end
+
+  def run_task
+    ENV["SOLR_CORE"] = "blacklight-core"
```
maxkadel: I'm worried that this is very different from how the other services connect to Solr, but maybe that's ok if we're going to refactor it?

spec/system/sanity_check_spec.rb
```diff
@@ -5,4 +5,14 @@
   it "runs rspec" do
     true
   end
+
+  describe 'index' do
```
maxkadel, line 9: Putting this here as a first pass was fine, but for the PR could you build out the test more? 
- Put it in a file with a name that aligns with what's being tested (if we have another system spec we don't have to keep the "sanity check" spec)
- Be sure that you are testing what you say you are testing. So, does this test prove that the data has been indexed to solr? Or does it prove that there is a static page that says the files have been indexed?

maxkadel (changes requested):

app/views/solr_task/run_task.html.erb
orangewolf: if its a flash message @maxkadel on what page would you display it? the one that has the button?

.env
```diff
@@ -20,7 +20,8 @@ CANTALOUPE_TAG=v1.0.0
 MANIFEST_TAG=v1.0.0
 
 # Solr
-SOLR_URL=http://solr:8983/solr
+SOLR_HOST=http://solr:8983/solr
```
orangewolf, line 23: these changes (all of the .env / .circle / Gemfile ones) do not appear to be related to this PR.  Why are they here and not in their own PR?

lib/tasks/index_sample_data.rake
```diff
@@ -9,9 +9,7 @@ namespace :yale do
 
   desc "Delete all solr documents"
   task clean_solr: :environment do
-    solr_core = ENV["SOLR_CORE"]
-    solr_url = ENV["SOLR_URL"] ||= "http://localhost:8983/solr"
-    solr = RSolr.connect url: "#{solr_url}/#{solr_core}"
+    solr = SolrService.connection
```
orangewolf, line 12: this is a great bit of cleanup

spec/system/sanity_check_spec.rb
orangewolf: I agree with Max here, though I'd take it a step further and question the existence of spec/system as a spec directory at all. Why does this exist? You are testing Rspec in this file... that is not what you came here to do ;-)

orangewolf (changes requested):
@maxkadel I think about half my questions are about @dylansalay code and about half are about yours. I think you may have pushed up to the wrong branch by accident?

.env
maxkadel: I'm moving them to their own PR. We were working on it in service of getting this PR passing CI, but I think they belong on their own branch. Working on it now.

spec/system/sanity_check_spec.rb
maxkadel: We usually use system specs for integration-type specs, or testing "from the outside in," using Capybara to drive a browser. There are some examples in Blacklight. The initial "sanity check" spec was just for testing the setup of Capybara and its integration with CircleCI, which is now working fine, so we can get rid of the "Rspec test" now that it's working.

app/views/solr_task/run_task.html.erb
maxkadel: That's what I was thinking (display it on the page with the button), but I also think this is fine for a first pass.

PR #41, opened 2020-06-05
Solr connection refactor

maxkadel:
* Cherry-pick the Solr refactor work so that it can be merged in its own PR
* Do not use dotenv-rails gem, since it is not behaving as expected in conjunction with docker / passenger / nginx
* Pin to slightly older version of Passenger Docker image to align with current Ruby version for our image (2.6.5)


maxkadel:
superseded

PR #42, opened 2020-06-05
Update base docker image and Ruby for latest passenger

maxkadel:


mlooney (approved):

Gemfile
```diff
@@ -3,9 +3,7 @@
 source 'https://rubygems.org'
 git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
-ruby '2.6.5'
-
-gem 'dotenv-rails', groups: [:development, :test]
+ruby '2.6.6'
```
orangewolf, line 6: I would argue that since Ruby is specified at the Dockerfile level and we only support running the app inside a container there is no need for this line to exist and that it is not dry

orangewolf (changes requested):

orangewolf:
well, I guess I was too slow on that! its not a big request but it would reduce the kind of lock up we saw earlier this week

Gemfile
mlooney: d'oh, fair point

PR #45, opened 2020-06-05
refactor solr connection

maxkadel:
Co-authored-by: martinlovell <martin.lovell@yale.edu>
Co-authored-by: dylansalay <dylansalay@gmail.com>
Co-authored-by: jpengst <jp@notch8.com>
Co-authored-by: kcompsci <lakeisha.robinson@yale.edu>
Co-authored-by: yjyale <yue.ji@yale.edu>


mlooney (changes requested):
see comment line 23, and later usage of the vars suggest it's a bug

app/services/solr_service.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class SolrService
+  def self.connection
+    solr_core = ENV["SOLR_CORE"] ||= "blacklight-core"
+    solr_host = ENV["SOLR_HOST"] ||= "http://localhost:8983/solr"
+    RSolr.connect url: "#{solr_host}/#{solr_core}"
```
mlooney: this will make ugly stuff if the "solr_host" value has an ending slash, consider using 

URI.join('http://foo/', '/bar/', '/baz').to_s

config/environments/test.rb
```diff
@@ -7,6 +7,7 @@
 
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
+  ENV["SOLR_CORE"] = "blacklight-test"
```
orangewolf, line 10: what do you think about?
`ENV['SOLR_CORE'] = ENV['SOLR_CORE_TEST'] || 'blacklight-test'`

so it can be customized and is 12factor

orangewolf:
one small thing and obv some red specs, but this is a step forward

app/services/solr_service.rb
maxkadel: Unfortunately this ends up leaving out '/bar/ in all the ways I've tried it.

mlooney: lol, I guess I should read the docs first

mlooney (approved):

PR #48, opened 2020-06-05
Voyager from mc

maxkadel:
Can now pull metadata from the Metadata Cloud from either Voyager (aka "ils"), or Ladybird.


maxkadel:
Superseded

PR #49, opened 2020-06-08
Allows us to refresh fixture data from the metadata cloud from ArchiveSpace, Voyager, or Ladybird

maxkadel:
* Now calls to the MetadataCloudService include which metadata_source should be used
* Ladybird is the source of authority for the relationships between Ladybird, Voyager, and ArchiveSpace records, so right now 
  * The MetadataCloudService grabs the oids from the CSV file, then
  * Opens the fixture file with the Ladybird record corresponding to that oid
  * From the fixture file determines the 
    * Voyager bib_id and barcode, if there is one, or
    * ArchiveSpace url 

Note, it looks like the MetadataCloud is providing odd info for the title field

Future refinement: We may want to create a model for "DigitalWork" or something similar, and save the oid, bib_id, barcode, and ArchiveSpace url to the database, and refresh periodically from Ladybird, possibly on a chron job or something.


maxkadel:
Changed to have 2 separate commits, one with relevant code changes, one with fixtures files, see PR#50

PR #50, opened 2020-06-08
Allows us to refresh fixture data from the metadata cloud from ArchiveSpace, Voyager, or Ladybird

maxkadel:
To see only relevant code changes, rather than fixture files, look at first commit. Second commit is only fixture updates.

* Now calls to the MetadataCloudService include which metadata_source should be used
* Ladybird is the source of authority for the relationships between Ladybird, Voyager, and ArchiveSpace records, so right now 
  * The MetadataCloudService grabs the oids from the CSV file, then
  * Opens the fixture file with the Ladybird record corresponding to that oid
  * From the fixture file determines the 
    * Voyager bib_id and barcode, if there is one, or
    * ArchiveSpace url 

Note, it looks like the MetadataCloud is providing odd info for the title field

Future refinement: We may want to create a model for "DigitalWork" or something similar, and save the oid, bib_id, barcode, and ArchiveSpace url to the database, and refresh periodically from Ladybird, possibly on a chron job or something.


app/services/metadata_cloud_service.rb
```diff
@@ -4,36 +4,78 @@
...
     metadata_cloud_username = ENV["MC_USER"]
     metadata_cloud_password = ENV["MC_PW"]
-    HTTP.basic_auth(user: metadata_cloud_username, pass: metadata_cloud_password).get(oid_url)
+    HTTP.basic_auth(user: metadata_cloud_username, pass: metadata_cloud_password).get(mc_url)
+  end
+
+  ##
+  # Takes an oid (Ladybird identifier) and a metadata source (allowed values are ladybird, ils, and aspace), and returns
+  # the appropriate URL to pull the metadata from the Yale Metadata Cloud
+  def build_metadata_cloud_url(oid, metadata_source)
```
bess, line 28: I'm going to merge this as-is but this method looks ripe for refactoring into three separate methods.

bess (approved):

PR #51, opened 2020-06-08
Re-worked Solr task button

dylansalay:
- Re-did work on Solr task button to incorporate docker image changes

### DIFF
--- a/spec/system/solr_tasks_spec.rb
+++ b/spec/system/solr_tasks_spec.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 require 'rails_helper'
 
-RSpec.describe "Solr Indexing Tasks", type: :system do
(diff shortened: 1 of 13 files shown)


spec/system/solr_tasks_spec.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+RSpec.describe "Solr Indexing Tasks", type: :system do
```
maxkadel: This test is looking great! 

Would you add "clean: true" info to this test? This makes it so that the Solr index is cleaned out before each "it" statement, so that we can be sure we don't get into an order-dependent test state with our Solr index. See the FixtureIndexingService spec for an example.

maxkadel (changes requested):

maxkadel (approved):

PR #53, opened 2020-06-08
Fixture indexing service takes metadata_source

maxkadel:
* Can now index to Solr from Voyager ("ils") or Ladybird.
* Updated UI buttons to allow for either

image


bess (approved):
🚀

PR #55, opened 2020-06-09
Ensure Bootstrap is added to asset pipeline correctly

maxkadel:
It looks like we may have added Bootstrap following a Rails 5 guide, we need to follow Rails 6 instructions, made changes based on this - https://dev.to/somnathpaul/add-bootstrap-4-to-your-ruby-on-rails-6-application-ole . Hopefully this will allow the asset pipeline to be created correctly in the deployed environments (currently not reflecting bootstrap styling) 

image

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -1,4 +1,3 @@
-@import "bootstrap";
 /*
  * This is a manifest file that'll be compiled into application.css, which will include all the files
  * listed below.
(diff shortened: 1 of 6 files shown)


maxkadel:
Deployed branch to yul-test3 cluster, appears to be working correctly
image

maxkadel:
... but not if you look at it via the load balancer  
image

fnibbit (approved):

PR #57, opened 2020-06-10
Service Status Display Table

dylansalay:
- Created table displaying container names and image tags


app/assets/stylesheets/application.scss
```diff
@@ -1,16 +1,6 @@
...
- * compiled file so the styles you add here take precedence over styles defined in any other CSS/SCSS
- * files in this directory. Styles in this file should be added after the last require_* statement.
- * It is generally better to create a new file per style scope.
- *
- *= require_tree .
- *= require_self
- */
+
+
+.service-container h3 {
```
maxkadel: See the latest updates to the bootstrap setup - I think you'll want to put any css in a "management" directory in app/assets/stylesheets and import it above bootstrap in management.scss (something like that). It should not go in this file - as you can see from the comment in the previous version of this file.

app/views/management/index.html.erb
```diff
@@ -10,3 +10,8 @@
     <%= link_to "Index Ladybird Records to Solr", metadata_source_path("ladybird"), class: "btn btn-primary" %>
     <%= link_to "Index ArchiveSpace Records to Solr", metadata_source_path("aspace"), class: "btn btn-primary" %>
 </div>
+
+<br /> 
+
+<%= render partial: "container" %>
```
maxkadel: Can we come up with something more descriptive for this partial?

ops/env.conf
```diff
@@ -1,9 +1,13 @@
 env TAG;
 env DB_ADAPTER;
+env MANAGEMENT_TAG
 env POSTGRES_HOST;
 env POSTGRES_DB;
 env POSTGRES_USER;
 env POSTGRES_PASSWORD;
 env PASSENGER_APP_ENV;
+env REGISTRY_HOST;
```
maxkadel: I'm not sure we're going to need the REGISTRY_HOST in this information, since I believe we're exclusively using dockerhub as our registry host. Same goes for above.

spec/view/management_spec.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+require 'rails_helper'
```
maxkadel, line 2: Nice test!

spec/view/management_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+
+RSpec.describe "Management", type: :feature do
+  describe 'index page' do
+    before do
+      ENV['REGISTRY_HOST'] = ''
+      ENV['REGISTRY_URI'] = 'yalelibraryit/dc-management'
+      ENV['MANAGEMENT_TAG'] = 'v1.2.3'
```
maxkadel: Since we're in the v1.x.x series for both Management and Solr right now, could we have these tags be something more obviously fake, so that we don't get confused? Like `v1Beta.Gamma` or something?

.circleci/config.yml
```diff
@@ -41,6 +41,11 @@ jobs:
           # Excludes tests that require Yale VPN access
           command:  |
             bundle exec rspec --tag ~vpn_only:true
+          environment:
```
maxkadel: Since these are being set explicitly in the test, I think it will only cause confusion to have them here.

app/views/management/_container.html.erb
```diff
@@ -0,0 +1,36 @@
...
+      <tr>
+        <td> Blacklight </td>
+        <td> </td>
+      </tr>
+      <tr>
+        <td> Solr </td>
+        <td> <%= "yalelibraryit/dc-solr:" + "#{ENV['SOLR_TAG']}" %> </td>
+      </tr>
+      <tr>
+        <td> iiif_image </td>
```
maxkadel: I think part of the difficulty of this ticket is that, currently, the deploy is being done using https://github.com/yalelibrary/yul-dc-camerata, and we'd want this page to show the versions that were in the latest Camerata deployment, not necessarily what is set in the management application itself.

Coordinating configuration between the applications is a tricky problem that is still in flux, but I see grappling with it as an integral part of this ticket. The fact that none of the non-Management tags are included suggests that maybe more thought could be put to that aspect of this problem.

maxkadel (changes requested):
This is a good start, but I'm concerned that the changes to circleci/config.yml and application.scss will not move us forward.

.circleci/config.yml
dylansalay: I removed env REGISTRY_HOST but I'm pretty sure we still need to set the other 3 here otherwise they will appear blank when the pipeline is running the specs

app/assets/stylesheets/application.scss
dylansalay: Addressed in latest commit

app/views/management/index.html.erb
dylansalay: Changed to `service_display`

ops/env.conf
dylansalay: Removed REGISTRY_HOST here

spec/view/management_spec.rb
dylansalay: Addressed in latest commit

app/assets/stylesheets/application.scss
maxkadel: I think you need to revert this file to how it is in master, currently.

maxkadel:
Could you resolve the edit conflicts and squash your commits, once you think this is ready to be reviewed again?

.circleci/config.yml
mark-dce: @orangewolf can you weight in based on your proposed changes to docker build and CI processes?  I definitely would not expect to see these values set here (and wouldn't think to look here without a lot of sleuthing)

Is there any way we can use the values in docker-compose.yml, .env, or .env.test in preference to putting them in the CircleCI config?

app/views/management/_service_display.html.erb
```diff
@@ -0,0 +1,36 @@
...
+  <table class="table table-bordered table-striped">
+    <thead class="thead-dark">
+      <tr>
+        <th scope="col"> Name </th>
+        <th scope="col"> Image Tag </th>
+      </tr>
+    </thead>
+      <tr>
+        <td> Management </td>
+        <td> <%= "#{ENV['REGISTRY_URI'] + ":" + ENV['MANAGEMENT_TAG']}" %> </td>
```
orangewolf: Lets try just using less syntax here
`<%= ENV['REGISTRY_URI'] %>:<%= ENV['MANAGEMENT_TAG']  %>`

.circleci/config.yml
orangewolf: We just need the code change below to allow the env vars to be blank w/o blowing up the code. This is good practice anyway so I think we can punt on the coming discussion about where env vars should be set so that they are not repeated so often.

ops/env.conf
```diff
@@ -1,9 +1,12 @@
 env TAG;
 env DB_ADAPTER;
+env MANAGEMENT_TAG
```
orangewolf: good catch - this is needed so that in production these variables get through to the running passenger processes. Otherwise they will always appear blank to the application causing this code to seem like it doesn't work (when it does but isn't getting the data it wants)

app/views/management/_container.html.erb
orangewolf: @maxkadel I'm not sure I understand what you mean here. This page reflects what is running in the container space at run time, not at build time.  Because the ENV vars are used, this page can simply report the available information from the running instance and therefor what is "up" at the given moment in time.  Camerata sets the correct info when it deploys to ECS or runs locally so this page should have what it needs when management is running alongside the other services. And when those services are not available, these other tags should be blank. If the names of those tags change then this code will need to change, but I don't really see a way around that.

.circleci/config.yml
dylansalay: @maxkadel @mark-dce @orangewolf 
The ENV variables have been removed from .circleci/config.yml.

maxkadel (commented):
re: my comment about not grappling with the deployed environment was based on the fact that this partial is not yet attempting to show the Database, Blacklight, or IIIF image server versions. I agree that it will still work, although I think maybe Camerata is using a "SERVICE_VERSION" rather than tag (d'oh).

app/views/management/_container.html.erb
maxkadel: re: my comment about not grappling with the deployed environment was based on the fact that this partial is not yet attempting to show the Database, Blacklight, or IIIF image server versions. I agree that it will still work, although I think maybe Camerata is using a "SERVICE_VERSION" rather than tag (d'oh).

maxkadel (approved):

maxkadel (approved):

PR #58, opened 2020-06-10
Create a model for ParentObjects and seed the database with example oids

maxkadel:
- The list of oids is the list that previously lived in the Fixture directory
- I'm not sure how "baked in" to make the seeding 
  - we could add it to the ops/nginx.sh for development & test? (add to where db:migrate is in the code)
    - Pro: automated
    - Con: not clear that it correctly seeds the test database (in my initial experiment)
    - Con: probably wouldn't want to do medium-term in production, and it could introduce dev / prod disparity
  - We could do it manually, as in the readme, 
    - Pro: Makes it transparent how the database is populated
    - Con: Having to manually seed the DB slows down development & test iterations slightly
    - Con:  Doesn't answer how we would populate the database in production.


maxkadel:
superseded by one with squashed commits

PR #75, opened 2020-06-16
Activity stream given date

maxkadel:


spec/lib/activity_stream_reader_spec.rb
```diff
@@ -80,11 +81,16 @@
...
-    #   expect(ActivityStreamLog.count).to eq 2
-    # end
+    it "can process the partial activity stream if there is a previous successful run" do
+      asl_old_success
+      expect(ActivityStreamLog.count).to eq 1
+      expect(ActivityStreamLog.last.object_count).to eq asl_old_success.object_count
+      asr.process_activity_stream
+      expect(ActivityStreamLog.count).to eq 2
+      # should be more than just one page of objects
+      expect(ActivityStreamLog.last.object_count).to be > 1000
```
bess: I'd like this to be a specific number. We're working with known files, so we should be getting the same answer here on each run, right? Let's put the specific number in so we don't get off-by-one errors creeping in anywhere.

bess (changes requested):
@maxkadel This looks great, but I am asking for one small change.

bess (approved):

PR #78, opened 2020-06-17
use same db configuration as blacklight

maxkadel:
Changes in support of running in production
- Use the same database username and password in production - matches Blacklight configuration
- Remove whitelisted hosts - allow any host for now in production
- Send / to management path (supports load balancer health checks)


mark-dce (approved):
👍

PR #103, opened 2020-06-23
Add temporary fix for Box pending more detailed requirements

maxkadel:
- Refactors building of solr document into its own method pending further refinements
Connected to https://github.com/yalelibrary/YUL-DC/issues/234


.rubocop_todo.yml
```diff
@@ -1,43 +1,29 @@
...
+# on 2020-06-23 16:21:15 +0000 using RuboCop version 0.63.1.
 # The point is for the user to remove these configuration records
 # one by one as the offenses are removed from the code base.
 # Note that changes in the inspected code, or installation of new
 # versions of RuboCop, may require this file to be generated again.
 
 # Offense count: 1
 Metrics/AbcSize:
-  Max: 60
+  Max: 54
```
orangewolf, line 11: this feels like an arbitrary rule change. is it intentional?

orangewolf (approved):

.rubocop_todo.yml
maxkadel: This is the auto-generated config. It has to do with adding the "or" statements to the `build_solr_document(id_prefix, oid, data_hash)` method. I'd pulled this out of the  `index_to_solr` method in hopes that it would simplify it enough to satisfy Rubocop, but it didn't.

I don't know enough about how we want to deal with field name differences to build a different method right now, but I think the refactor of the solr document build is worthwhile to get in. 

We could only map the Ladybird value for `box_ssim` now and not have to raise this metric?

PR #105, opened 2020-06-23
Add visibility from SOA Ladybird for all items

maxkadel:
- Allows records from any source to be visible in Discovery
- We need to treat Ladybird as the source of authority for visibility for items, even if the record we're showing is from Voyager or ArchiveSpace. 

- QUESTIONS:
  - I have concerns about performance at scale with both pushing to Solr and doing a database query (albeit one that should currently be pretty fast)
- I know we've talked about not *combining* data from different sources of authority, and this kinda starts us down this road, maybe? 

Top record is Voyager, bottom record is Ladybird.
image

Connected to https://github.com/yalelibrary/YUL-DC/issues/274
Connected to https://github.com/yalelibrary/YUL-DC/issues/260


app/services/metadata_cloud_service.rb
```diff
@@ -63,6 +64,7 @@ def find_source_ids_for(oid)
       aspace_uri: aspace_uri,
       holding: holding,
       item: item,
+      visibility: visibility,
```
orangewolf: I'd lean toward doing the mapping right here since it's short instead of the more c like style w/ the extra variable.

bess (approved):
Looks good to me, but I won't merge yet in case @orangewolf wants to weigh in first.

app/services/fixture_indexing_service.rb
```diff
@@ -62,10 +62,7 @@ def build_solr_document(id_prefix, oid, data_hash)
       oid_ssim: data_hash["oid"] || oid,
       identifierMfhd_ssim: data_hash["identifierMfhd"],
       identifierShelfMark_ssim: data_hash["identifierShelfMark"],
-      # I do not think the current box_ssim is how we want to continue to do deal with differences in field names
-      # However I do not think we currently have enough information to create the alternative (Max)
-      # Ladybird data_hash["box"] || Voyager data_hash["volumeEnumeration"] || ArchiveSpace data_hash["containerGrouping"]
-      box_ssim: data_hash["box"] || data_hash["volumeEnumeration"] || data_hash["containerGrouping"],
+      box_ssim: calculate_box_ssim(data_hash),
```
mark-dce: FWIW - as a math major, I read the word "calculate" and expect the method to be doing some kind of arithmetic computation.  For this kind of thing, I might use names like "extract" "read" "parse".

orangewolf (approved):

PR #114, opened 2020-06-25
Refactor Metadata cloud service - put parsing tasks in own class

maxkadel:
- Change from instance methods to class methods for classes that currently don't have a state needed (MetadataCloudService, FixtureIndexingService, and FixtureParsingService).
- Split out methods for file parsing from MetadataCloudService into a new FixtureParsingService class
  - Update calls to appropriate methods / classes

Connected to https://github.com/yalelibrary/YUL-DC/issues/221


app/services/fixture_parsing_service.rb
```diff
@@ -0,0 +1,65 @@
...
+  end
+
+  ##
+  # Takes a csv file and returns an array containing the values from the first column
+  def self.build_oid_array(oid_path)
+    fixture_ids_table = CSV.read(oid_path, headers: true)
+    fixture_ids_table.by_col[0]
+  end
+
+  def file_prefix(metadata_source)
```
mlooney: can this be static?  if so, you wouldn't need to instantiate this class above.

app/services/fixture_indexing_service.rb
```diff
@@ -95,11 +94,11 @@ def extract_box_ssim(data_hash)
   end
 
   def self.index_to_solr(oid, metadata_source)
-    mcs = MetadataCloudService.new
+    fps = FixtureParsingService.new
```
mlooney: see below

app/services/fixture_parsing_service.rb
```diff
@@ -0,0 +1,65 @@
...
+  def self.find_source_ids
+    fps = FixtureParsingService.new
+    parent_objects = ParentObject.all
+    parent_objects.each do |parent_object|
+      oid = parent_object["oid"]
+      fps.find_source_ids_for(oid)
+    end
+  end
+
+  def find_source_ids_for(oid)
```
mlooney: can this be static?

app/services/metadata_cloud_service.rb
```diff
@@ -1,5 +1,4 @@
 # frozen_string_literal: true
-require 'csv'
 
 class MetadataCloudService
```
mlooney, line 3: same deal for this guy, it looks to me like it has no state, if you make everything static, you won't need to instantiate

mlooney (approved):

PR #117, opened 2020-06-26
Add dependent object model method to populate

maxkadel:
* Currently, ArchiveSpace records require use of these dependent URIs in order to retrieve relevant updates from the ActivityStream. 
  * The MetadataCloud now returns dependent URIs, which represent the URIs for child, parent, or sibling objects to our "object of interest", aka ParentObject. 
* This PR creates a model for DependentObjects, and the relationship between them and ParentObjects, using OIDs as the unique identifier
* The ActivityStreamLog now includes both how many items were in the relevant portion of the Activity Stream, and how many records were updated based on that run of the Activity Stream.


app/services/fixture_parsing_service.rb
```diff
@@ -6,7 +6,7 @@ class FixtureParsingService
   # Takes an oid and a metadata_source and returns a hash of the fixture file associated with that oid and metadata_source
   def self.fixture_file_to_hash(oid, metadata_source)
     fixture_file_folder = Rails.root.join("spec", "fixtures", metadata_source)
-    file_prefix = FixtureParsingService.file_prefix(metadata_source)
+    file_prefix = file_prefix(metadata_source)
```
orangewolf: naming a variable exactly what the method is named can be confusing at best, and mis-read by Ruby in some cases

orangewolf: (yes, i see that all we did was pull the class prefix off here, but it made the fact that this was gross more obvious ;-)

app/lib/activity_stream_reader.rb
```diff
@@ -56,6 +56,7 @@ def find_by_id(item)
     elsif metadata_source == "aspace"
       source_id = ("/" + source_id_type + "/" + source_id).to_s
       oid = ParentObject.where("aspace_uri" => source_id.to_s)&.first&.oid
+      oid = DependentObject.where(dependent_uri: source_id.to_s)&.first&.parent_object_id&.to_s if oid.nil?
```
orangewolf: this feels like maybe a method that's missing. that's a _lot_ of nil checking in the middle of doing find_by_id. (6 in 2 lines by my count). Maybe extracting it out to a `oid_from_source_id(source_id)` method could make that more readable / reusable?

.env
```diff
@@ -17,7 +17,7 @@ SOLR_VERSION=v1.0.0
 IIIF_IMAGE_VERSION=v1.0.0
 POSTGRES_VERSION=v1.0.0
 BLACKLIGHT_VERSION=v1.6.0
-MANAGEMENT_VERSION=v1.7.2
+MANAGEMENT_VERSION=v1.8.0
```
orangewolf: is it generally the practice to bump the version while doing code changes or do we release separately? curious what the pattern has been here

maxkadel: I think the pattern has not been consistent, but I think doing it when we do the release would be better. Maybe we could add it to the readme on deployment in a separate PR, I can take it out of this PR.

martinlovell:
I wonder if it would be good to add metadata_source to the DependentObject model.
Then you could reduce `ActivityStreamReader.find_item_id()` to:
```
  def find_by_id(item)
    parsed_identifier_uri = parse_identifier_uri(item)
    metadata_source = parsed_identifier_uri[0]
    source_id_type = parsed_identifier_uri[1]
    source_id = parsed_identifier_uri[2]
    dependent_object = DependentObject.where(dependent_uri: source_id.to_s)&.first
    return false unless dependent_object
    oids_for_update.add([dependent_object.parent_object_id, dependent_object.metadata_source])
    true
  end
  ```

This way all sources are treated the same in relationship to the ActivityStream.

(`FixtureParsingService.self.find_dependent_uris()` would need to be updated to include the metadata source with the DependentObject.)

(This code only gets the first dependent_object, but it should really get all of them and add all to oids_for_update.    Please see comment about using .first above.

To handle all dependent objects, maybe something like this (not tested at all):
```
  def find_by_id(item)
    parsed_identifier_uri = parse_identifier_uri(item)
    metadata_source = parsed_identifier_uri[0]
    source_id_type = parsed_identifier_uri[1]
    source_id = parsed_identifier_uri[2]
    dependent_objects = DependentObject.where(dependent_uri: source_id.to_s)
    return false unless dependent_objects && dependent_objects.count > 0
    dependent_objects.each do |dependent_object|
      oids_for_update.add([dependent_object.parent_object_id, dependent_object.metadata_source])
    end
    true
  end
```
)

File: app/lib/activity_stream_reader.rb (hunk 2 above)
Comment (martinlovell) [reply app/lib/activity_stream_reader.rb]: It's possible (and likely) that multiple aspace objects may depend on one entry in the activity stream.
So, just getting the first one will not find all the oids that need to be updated in some cases.

File: app/lib/activity_stream_reader.rb [hunk 4]
```diff
@@ -47,20 +47,9 @@ def relevant?(item)
...
-      oids_for_update.add([oid, metadata_source])
-    elsif metadata_source == "aspace"
-      source_id = ("/" + source_id_type + "/" + source_id).to_s
-      oid = ParentObject.where("aspace_uri" => source_id.to_s)&.first&.oid
-      return false unless oid
-      oids_for_update.add([oid, metadata_source])
-    else
-      return false
-    end
+    oid = DependentObject.where(dependent_uri: parse_dependent_uri(item))&.first&.parent_object_id&.to_s
```
orangewolf: find_by_id makes me think:
1) I pass in an id (not an item)
2) I get back either an object or nil

this is mostly because it is such an "active recordy" name. That doesn't seem to be what this method does at this point, it returns true or false now.

also it seems like this still may need to be a loop according to @martinlovell comments? everything else looks good to me now.

orangewolf (commented):
everything else looks solid, still focused on the find_by_id method a bit

app/services/fixture_parsing_service.rb
```diff
@@ -43,6 +42,18 @@ def self.find_source_ids_for(oid)
...
+    hash = fixture_file_to_hash(oid, metadata_source)
+    hash["dependentUris"].each do |uri|
+      dep_obj = DependentObject.find_or_create_by(
+        dependent_uri: uri,
+        metadata_source: metadata_source,
+        parent_object_id: oid
+      )
+      dep_obj.save
+    end
+  end
```
orangewolf, line 55: 👍

orangewolf (approved):

orangewolf:
@maxkadel this looks good to me, but we really should have Martin's eyes too since the relationship part is so complex and essential

app/lib/activity_stream_reader.rb
```diff
@@ -26,62 +36,59 @@ def process_page(page_url)
     process_page(previous_page_link(page)) if previous_page_link(page)
   end
 
+  def process_item(item)
+    dependent_uri = parse_item_identifier(item)
+    dependent_objects = dependent_objects_based_on(dependent_uri)
+    return false unless dependent_objects
+    add_to_parent_objects_for_update_set(dependent_objects)
+    @tally += 1
```
martinlovell: It may be that @tally is off since it might be more than one parent.  

It's right if object_count is the number items with one or more document matches, but is seems like object_count might mean the total number of documents that were updated.

If it's the number of documents that were updated, it might be possible to get rid of tally and just use parent_objects_for_update.count for the log.object_count in process_activity_stream()

PR #118, opened 2020-06-29
Add more tasks to UI dashboard

maxkadel:
- Updates ParentObjects with identifiers & visibility in the database based on json fixture files from MetadataCloud
- Workaround for current lack of DB persistence in production-like environments:
  - If there are no ParentObjects in the database, adds them from the fixture file
- Run ActivityStreamReader.update from the UI dashboard
  - I doubt this will run in production, due to lack of VPN access, but could be good step forward for 
        a. demo-ing it in development environments, 
        b. demonstrating that a solution for connection for production-like environments is needed
- Updates flash message to reflect current behavior for Solr Indexing jobs

image


maxkadel:
Unclear if this matches requirements, draft for now.

PR #124, opened 2020-07-01
257 oid csv upload

jpengst:


app/controllers/oid_imports_controller.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+class OidImportsController < ApplicationController
+  rescue_from ActiveRecord::RecordInvalid, with: :show_errors
```
orangewolf: I don't think this is used?

app/controllers/oid_imports_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+    end
+  end
+
+  private
+
+    def oid_import_params
+      params.require(:oid_import).permit(:file)
+    end
+
+    def show_errors(exception)
```
orangewolf: if the top line isn't used, then this is not either.

orangewolf (approved):

PR #137, opened 2020-07-08
Add an OID minter service and controller

mikeapp:
Pull request for yalelibrary/YUL-DC#320 and yalelibrary/YUL-DC#321.   Detailed requirements are linked from the first issue.

**Testing**

* Delivers a single OID as text: http://localhost:3001/management/api/oid/new.txt
* Delivers a single OID as JSON: http://localhost:3001/management/api/oid/new.json
* Delivers multiple OIDs as text: http://localhost:3001/management/api/oid/new/5.txt
* Delivers multiple OIDs as JSON: http://localhost:3001/management/api/oid/new/5.json
* Returns an error with non-integer parameter or to requests for other content types

**Issues**

* The options for initializing a DB-specific construct are not great, see https://stackoverflow.com/questions/9144556/create-sequence-in-migration-not-reflected-in-schema.  It was easiest to use SQL 'CREATE ... IF NOT EXISTS ...' on each call to ensure that the sequence is initialized.  The overhead is 1-2ms.

* The initial value for the sequence is set in the configuration for the environment.  Development and test have very high starting values so it will be clear from visual inspection whether the value came from dev, test, or production.   I'm not sure whether there would be any value in moving this parameter to an environment variable so that it can be set via Docker.


app/services/oid_minter_service.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+#
+class OidMinterService
+  def self.generate_oids(number)
+    oids = []
+    initialize_sequence!
+    number.times do
```
maxkadel, line 7: In order to "minimize large gaps in sequence" - do we want to limit the number of oids a service is allowed to request at once? (scenario: a service accidentally requests 1,000 oids but only actually needs 10)

maxkadel (approved):

app/services/oid_minter_service.rb
mikeapp: I agree but wasn't sure what the max should be.  100 might be reasonable, our largest object is ~2,500 images.  This would be good for another ticket.

app/services/oid_minter_service.rb
```diff
@@ -0,0 +1,19 @@
...
+      result = ActiveRecord::Base.connection.execute("SELECT NEXTVAL('OID_SEQUENCE')")
+      oids.push result.getvalue(0, 0)
+    end
+    oids
+  end
+
+  def self.initialize_sequence!
```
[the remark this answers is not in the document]
mikeapp: I thought about the rake task or other one-time setup, but wasn't sure how that would impact other automation.  I would defer this and logging to another ticket.

PR #138, opened 2020-07-08
Sample and save

dylansalay:
Changes allow us to produce CSV containing a list of fields and display various attributes from a random sampling of ~10,000 public items from LadyBird and from the local file system.


app/services/metadata_sampling_service.rb
```diff
@@ -48,6 +107,20 @@ def self.record_initial_findings(metadata_sample)
     end
   end
 
+  def self.reverse_engineer_oids
```
mark-dce: would it be clearer to call this "extract_oids_from_filenames"

mark-dce (approved):

PR #141, opened 2020-07-10
322 Authenticate oid request

dylansalay:
Protects OID Minter Service using HTTP Basic Authentication

Screen Shot 2020-07-10 at 3 38 00 PM

### DIFF
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -79,7 +79,7 @@ Devise.setup do |config|
   # enable this with :database unless you are using a custom strategy.
   # The supported strategies are:
   # :database      = Support basic authentication with authentication key + password
-  # config.http_authenticatable = false
+  config.http_authenticatable = true
 
   # If 401 status code should be returned for AJAX requests. True by default.
   # config.http_authenticatable_on_xhr = true
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -1,0 +1,311 @@
+# frozen_string_literal: true
+
+# Assuming you have not yet modified this file, each configuration option below
+# is set to its default value. Note that some are commented out while others
+# are not: uncommented lines are intended to protect your configuration from
+# breaking changes in upgrades (i.e., in the event that future versions of
+# Devise change the default values for those options).
+#
+# Use this hook to configure devise mailer, warden hooks and so forth.
+# Many of these configuration options can be set straight in your model.
+Devise.setup do |config|
+  # The secret key used by Devise. Devise uses this key to generate
+  # random tokens. Changing this key will render invalid all existing
+  # confirmation, reset password and unlock tokens in the database.
+  # Devise will use the `secret_key_base` as its `secret_key`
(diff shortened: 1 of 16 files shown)


config/initializers/devise.rb
```diff
@@ -0,0 +1,311 @@
...
+
+  # ==> Controller configuration
+  # Configure the parent class to the devise controllers.
+  # config.parent_controller = 'DeviseController'
+
+  # ==> Mailer Configuration
+  # Configure the e-mail address which will be shown in Devise::Mailer,
+  # note that it will be overwritten if you use your own mailer class
+  # with default "from" parameter.
+  config.mailer_sender = 'please-change-me-at-config-initializers-devise@example.com'
```
maxkadel, line 27: if we have this un-commented, should we make this a real email address? or should we comment it out?

db/migrate/20200709205615_devise_create_users.rb
```diff
@@ -0,0 +1,44 @@
...
+      ## Lockable
+      # t.integer  :failed_attempts, default: 0, null: false # Only if lock strategy is :failed_attempts
+      # t.string   :unlock_token # Only if unlock strategy is :email or :both
+      # t.datetime :locked_at
+
+
+      t.timestamps null: false
+    end
+
+    add_index :users, :email,                unique: true
```
maxkadel, line 39: I know this is probably just all the defaults, but do we really want to use email? or do we want to use username? I think if we do want to use username it'd be fine to do it on another pass.

spec/requests/oid_minter_request_spec.rb
```diff
@@ -2,69 +2,93 @@
...
-      expect(json_response['oids']).not_to be_nil
-      oids = json_response['oids']
-      expect(oids).to be_a Array
-      expect(oids.length).to be 1
-      expect(oids[0]).to be_a Integer
+  let(:user) { FactoryBot.create(:user) }
+  describe "when authenticated" do
+    context "Requesting a single OID" do
+      it "returns a successful JSON response" do
+        headers = login_header(user).merge('ACCEPT' => "application/json")
```
maxkadel, line 9: I wasn't familiar with this `login_header(user)` method, that seems super handy.

bess (approved):

config/initializers/devise.rb
dylansalay: I think it makes sense to comment it out for now. Maybe we add an email if we choose to implement devise authentication for the application?

db/migrate/20200709205615_devise_create_users.rb
dylansalay: Yes, we just went with the defaults. But we can discuss and change to username if that is a better solution.

spec/requests/oid_minter_request_spec.rb
dylansalay: It's not a Rails built in helper method - just a custom method we created in spec/support/auth_helper.rb to handle the authentication login.

PR #142, opened 2020-07-11
Configure yarn correctly, set up volumes to boost dev speed

orangewolf:
Node modules were not being installed properly, preventing webpacker from being present. If yarn install ever got run, it would be fine and Rails would update it as need it, but it has to happen one time so that webpack exists before it will run again.

Because I was configuring yarn / node_modules I brought in some volume logic which massively speeds up asset generation in Rails+Docker+Mac (or Windows).  It won't change how things work for Linux users, but it prevents an exponential slow down on other OS. (9s to 3s now for the little tiny set we have so far).  

I'll make sure this same change gets sync'd Blacklight during the Camerata work I'm doing this week.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,7 +10,9 @@ services:
     ports:
       - '3001:3001'
     volumes:
-      - .:/home/app/webapp
+      - .:/home/app/webapp:cached
+      - node_modules:/home/app/webapp/node_modules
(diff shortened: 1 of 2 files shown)


orangewolf:
Mark - it is worth noting that until the master docker image is rebuild with this fix in place, the pull / up process in the readme will continue to exhibit this webpack behavior.

fnibbit (approved):

PR #144, opened 2020-07-13
Refactor parent objects

maxkadel:


maxkadel:
I'm marking this as a draft because I don't think it's ready to merge (there's a lot of duplication of the MetadataCloudService)

app/controllers/management_controller.rb
```diff
@@ -6,7 +6,7 @@ def index
   end
 
   def index_to_solr
-    @metadata_source = FixtureIndexingService.index_fixture_data(params[:metadata_source])
+    @ms = FixtureIndexingService.index_fixture_data(params[:ms])
```
orangewolf: I don't feel like this is better. are we trying to appease rubocop or something?

app/controllers/parent_objects_controller.rb
```diff
@@ -0,0 +1,148 @@
...
+
+  # GET /parent_objects/new
+  def new
+    @parent_object = ParentObject.new
+  end
+
+  # GET /parent_objects/1/edit
+  def edit; end
+
+  def build_metadata_cloud_url(metadata_source, lb_record = "")
```
orangewolf: this should be private or maybe a helper?

app/controllers/parent_objects_controller.rb
```diff
@@ -0,0 +1,148 @@
...
+  end
+
+  def mc_get(mc_url)
+    metadata_cloud_username = ENV["MC_USER"]
+    metadata_cloud_password = ENV["MC_PW"]
+    HTTP.basic_auth(user: metadata_cloud_username, pass: metadata_cloud_password).get(mc_url)
+  end
+
+  # Takes a JSON record from the MetadataCloud and saves the Ladybird-specific info to the DB
+  def save_ladybird_info_to_db(lb_record)
```
orangewolf: this should be inverted to be @parent_object.lb_record = lb_record. all this logic can then live in lb_record=(value)

app/controllers/parent_objects_controller.rb
```diff
@@ -0,0 +1,148 @@
...
+    when "aspace"
+      "https://metadata-api-test.library.yale.edu/metadatacloud/api/aspace#{lb_record['archiveSpaceUri']}"
+    end
+  end
+
+  def mc_get(mc_url)
+    metadata_cloud_username = ENV["MC_USER"]
+    metadata_cloud_password = ENV["MC_PW"]
+    HTTP.basic_auth(user: metadata_cloud_username, pass: metadata_cloud_password).get(mc_url)
+  end
```
orangewolf: mc_get your mc_griddle in your mc_morning. we'll want to move this down to the model level or a service.

app/controllers/parent_objects_controller.rb
```diff
@@ -0,0 +1,148 @@
...
+  def initial_mc_data
+    lb_record = fetch_record("ladybird")
+    save_ladybird_info_to_db(lb_record)
+    v_record = fetch_record("ils", lb_record)
+    save_voyager_info_to_db(v_record)
+    return unless lb_record["archiveSpaceUri"]
+    a_record = fetch_record("aspace", lb_record)
+    save_aspace_info_to_db(a_record)
+  end
+
```
orangewolf: from the above comment to here is all "in the model" in my mind

app/controllers/parent_objects_controller.rb
```diff
@@ -0,0 +1,148 @@
...
+  # POST /parent_objects
+  # POST /parent_objects.json
+  def create
+    @parent_object = ParentObject.new(parent_object_params)
+    initial_mc_data
+    respond_to do |format|
+      if @parent_object.save
```
orangewolf, line 29: once we refactor the above code, this is primed and ready to handle the mc_data and anything in the future we might throw at it.

app/views/management/index.html.erb
```diff
@@ -6,11 +6,11 @@
     <button type="button" class="btn btn-primary dropdown-toggle mr-3" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
       Index Metadata to Solr
     </button>
-    
+
     <div class="dropdown-menu">
-      <%= link_to "Index Voyager Records to Solr", metadata_source_path("ils"), class: "dropdown-item" %>
-      <%= link_to "Index Ladybird Records to Solr", metadata_source_path("ladybird"), class: "dropdown-item" %>
-      <%= link_to "Index ArchiveSpace Records to Solr", metadata_source_path("aspace"), class: "dropdown-item" %>
+      <%= link_to "Index Voyager Records to Solr", ms_path("ils"), class: "dropdown-item" %>
```
orangewolf: why ms?

app/controllers/parent_objects_controller.rb
orangewolf: on second thought if metadata_source is a model now. it should now how to make these urls for us.

orangewolf (commented):
I can totally see where you are headed here Max and it feels like the right direction. I think if we go a little further the clouds will clear up and all of a sudden it will look really nice.

app/controllers/management_controller.rb
maxkadel: Agree this is ugly - it was to get it out of the way so that I could turn MetadataSource into its own model, and this was blocking the creation of the routes

I'll see whether, now that MetadataSource is its own model, this can be refactored.

app/views/management/index.html.erb
maxkadel: Oh, this is because of trying to address the same issue as your first comment - conflicting routes when trying to create a new model. Can hopefully come up with a better solution.

spec/models/metadata_source_spec.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe MetadataSource, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
orangewolf, line 6: Id normally ask for this to go, but given the follow on ticket of "add more specs" this seems fine

orangewolf (approved):
Looks really solid @maxkadel

PR #148, opened 2020-07-15
Test coverage badge

jpengst:
testing changes


maxkadel:
Since we hard coded the coveralls api key, which should probably stay private, will deleting the branch sufficiently remove it from the git history, or do we need to do anything additional?

PR #150, opened 2020-07-16
Goobi parent objects

maxkadel:
Prepare for Goobi upload

- Add defaults for ParentObjects (visibility, reading order, and pagination)
- ParentObjects only need to pull from MetadataCloud for their Authoritative Metadata Source (aka SOA), and from Ladybird in order to get any pre-requisite identifiers.
  - Update form to reflect this change
- Only OID and Authoritative Metadata Source are needed to create a ParentObject right now, remove other items from form to discourage them being editable without business needs being clear
- Expand ParentObject model & system tests


app/models/concerns/mets_structure.rb
```diff
@@ -0,0 +1,95 @@
+# frozen_string_literal: true
+
+module MetsStructure
+  # def structure
+  #   structure_type('logical')
+  # end
+  #
+  # def structure_for_volume(volume_id)
+  #   volume = volume_nodes.find { |vol| vol.attribute("ID").value == volume_id }
```
mark-dce: where does the term "volume" come from?  I wonder if we want something more generic or specifically from METS here?  I think "volume" might carry semantic weight that we don't want here - i.e. a volume for Beinecke is one specific type of artifact and we'll have lots of different things being described.

app/models/concerns/mets_structure.rb
```diff
@@ -0,0 +1,95 @@
...
+  #
+  #   def in_scope(node)
+  #     if multi_volume?
+  #       node.parent.parent.name == 'div'
+  #     else
+  #       node.parent.name == 'div'
+  #     end
+  #   end
+  #
+  #   def label_from_related_objects(id)
```
mark-dce: I don't see examples of related objects in our samples - where does this come from - are we pulling in dead code that we don't need?

app/models/mets_document.rb
```diff
@@ -0,0 +1,126 @@
...
+  def initialize(mets_file)
+    @source_file = mets_file
+    @mets = File.open(@source_file) { |f| Nokogiri::XML(f) }
+  end
+
+  def oid
+    @mets.xpath("//goobi:metadata[@name='CatalogIDDigital']").first.content.to_s
+  end
+
+  # unused
```
mark-dce: This looks like a lot of cut and paste code - folks working on this later won't know why it's here and whether is safe to delete or part of a feature that got abandoned, or ???  I'm not sure we want to pull it in if it doesn't help give context - I'd rather see a link to the source where we cribbed this from that says - this code was inspired by ... see (link) for more.

spec/models/parent_object_spec.rb
```diff
@@ -3,5 +3,63 @@
...
-  pending "add some examples to (or delete) #{__FILE__}"
+  before do
+    prep_metadata_call
+  end
+  context "a newly created ParentObject with just the oid and default authoritative_metadata_source (Ladybird for now)" do
+    let(:po) { described_class.create(oid: "2004628") }
+    before do
+      stub_request(:get, "https://metadata-api-test.library.yale.edu/metadatacloud/api/ladybird/oid/2004628")
+        .to_return(status: 200, body: File.open(File.join(fixture_path, "ladybird", "2004628.json")).read)
+      po
```
mark-dce: The stubbing is cool - you should walk through this for everyone.

I'm not sure I understand the value of separating the let and the call to `po` instead of just having
`po = described_class.create(oid: "2004628")` in the before block.  I.E. a line that just says `po` is pretty opaque - at least maybe call it `parent_object` there's really no tax on extra keystrokes and they make the code much more legible in the future.

spec/models/parent_object_spec.rb
```diff
@@ -3,5 +3,63 @@
...
+    end
+
+    it "pulls from the MetadataCloud for Ladybird and not Voyager or ArchiveSpace" do
+      expect(po.authoritative_metadata_source_id).to eq 1 # 1 is Ladybird
```
mark-dce: This might be easier for someone else to quickly parse if it read  
```
LADYBIRD = 1
expect(parent_object.authoritative_metadata_source_id).to eq LADYBIRD
```

If we end up referencing 1 / 2 / 3 for LadyBird, ASpace, Voyager in lots of places, it might even make sense to set up global constants.  OR we could ask the metadata cloud to pass a string ID that came from a semantic vocabulary ["LadyBird", "Voyager", "ASpace"]

app/models/concerns/mets_structure.rb
maxkadel: Yeah, I'm not sure how much of this we're going to want, that's why Rob suggested commenting it out. Maybe linking would be better.

app/views/parent_objects/_form.html.erb
```diff
@@ -16,54 +16,9 @@
     <%= form.text_field :oid %>
   </div>
 
-  <div class="field">
```
orangewolf, line 19: Does this means that there is now no view (show or edit) of these fields?. That seems like step backwards to me.  Maybe leave this partial alone stop having the new.html.erb reference it (so new and edit are independent for now)?

orangewolf (approved):
Seems overall a good change set. my only question / concern was around the view and if you want to override that feel free.

app/views/parent_objects/_form.html.erb
maxkadel: I think maybe we need more info on requirements, etc., so I'm reverting this change so we can have more discussion about what we want these to look like.

PR #151, opened 2020-07-16
OID request logger and spec

dylansalay:
Requests to the generate OID's via the OID minter service log the:

- User email
- IP Address
- OID's generated
 
Screen Shot 2020-07-16 at 1 27 19 PM


martinlovell (approved):
Very nice test.

PR #160, opened 2020-07-21
Seed production and dev on load, run OID Minter on seed

orangewolf:


db/seeds.rb
```diff
@@ -5,6 +5,8 @@
 #
 #   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
 #   Character.create(name: 'Luke', movie: movies.first)
+OidMinterService.initialize_sequence!
+puts "Oid Minter Initialized"
```
maxkadel: Would it be helpful to output what number the Oid Minter initialized at? Seems like a good first-step sanity check.

maxkadel (changes requested):

db/seeds.rb
maxkadel: Thanks, looks good.

maxkadel (approved):
Approving, Mark has asked to have a larger convo about always seeding the database in development, since it requires the VPN to be open.

PR #161, opened 2020-07-21
seeds will not error when ran a second time

jpengst:
Guard clause to prohibit ParentObject creation if seeds have already been ran.


db/seeds.rb
```diff
@@ -55,11 +55,13 @@
   fixture_ids_table = CSV.read(oid_path, headers: true)
   oids = fixture_ids_table.by_col[0]
 
-  oids.each do |row|
-    po = ParentObject.new
-    po.oid = row
-    po.save
-    puts "#{po.oid} saved"
+  unless ParentObject.count > 0
```
maxkadel: My understanding for this ticket was that this was supposed to follow the pattern for the other seeds. 

If we did that, a developer could add an example oid or two to the CSV, re-run the seed file, and have those in their local environment, which I could see as being really helpful for fast development.

maxkadel (changes requested):

maxkadel (approved):

PR #163, opened 2020-07-22
Refactor metadata cloud service

maxkadel:
- Uploading a CSV now creates a ParentObject in the database for all rows in the CSV
- Prepare for same functionality for METs xml uploads

Currently, the ActivityStreamReader needs the save-to-json methods in the MetadataCloudService, will refactor this code in a future PR.


app/models/parent_object.rb
```diff
@@ -23,6 +24,17 @@ def default_fetch
     end
   end
 
+  def authoritative_json
```
mark-dce, line 27: It seems like you're writing a `.to_json` method - is there a reason not to just override the default `.to_json` method here?  (i.e. is there ever a reason you'd want the output from just `parent_object.to_json` instead of `parent_object.authoritative_json` - if not, I'd be inclined to just override the default `.to_json` method.

mark-dce: Also, this will silently fail (i.e. just return nil) if the metadata_cloud_name is not in the expected list.  You should probably raise an error if none of the cases match
```
when 
   ... 
else
    raise "Unexpected metadata cloud name: #{authoritative_metadata_source.metadata_cloud_name}"
end

**Review by mark-dce — CHANGES_REQUESTED** [review]

File: app/models/parent_object.rb (hunk 1 above)
Comment (orangewolf) [reply app/models/parent_object.rb:27]: I agree about the raise, that's a good catch. We're not overriding to_json since this doesn't write the active record fields to json, but instead selects which of the jsonb fields is the one that should be used to write out to the file. In my opinion to_json is about the object as json, this is about one fields inside the object that happens to be json.

File: app/models/concerns/json_file.rb [hunk 2]
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module JsonFile
```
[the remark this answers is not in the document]
orangewolf: we know we want it for child object so it seemed like a reasonable time to work on the pattern. however that in and of itself would be a premature optimization.  the other reason to separating it in to a concern is that though it is related to the primary function of the parent object, it is not in and of itself part of that function. By making it a separate concern we can keep the parent object complexity down w/o resorting to more complex structural paradigms like service objects.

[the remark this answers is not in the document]
orangewolf: (its also not a big deal to move it into the object for now)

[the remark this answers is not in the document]
mark-dce: If there's a reason for it, I'm fine with it.  That reason wasn't apparent in the PR itself.  So I'm also good with either solution.

app/models/parent_object.rb
mark-dce: I'm probably conflating this with the `.to_solr` pattern for old Hydra-based apps - i.e. give me the thing serialized in a way that I can index it to Solr.

mark-dce (approved):

PR #166, opened 2020-07-22
Logging

mlooney:


config/application.rb
```diff
@@ -27,6 +27,11 @@ class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
     config.load_defaults 6.0
 
+    config.log_level = :debug
+    STDOUT.sync = true # turn off log buffering
```
orangewolf, line 31: reposting question from yesterday: at first glance this seems to make all logging requests synchronous writes with the logger. This means IO interrupts during request processing and would be potentially very bad for performance.  Do these writes get buffered somewhere else or is there some other mitigating factor?

ops/webapp.conf
```diff
@@ -1,4 +1,7 @@
+passenger_log_file /dev/stdout;
```
orangewolf, line 1: did you remove or disable the log tailer? passenger images currently runs a process that tees the logs in to stdout, which we should disable if we are going about this directly instead.

orangewolf (changes requested):

orangewolf (approved):

PR #168, opened 2020-07-23
METs upload creates parent object

maxkadel:
- METs upload creates a ParentObject
- Goobi XML Uploader validates that
  - xml file contains an OID
  - xml file contains METs xml
  - all image files are present


app/models/mets_document.rb
```diff
@@ -26,6 +20,11 @@ def valid_mets?
     true
   end
 
+  def all_images_present?
+    image_present = files.map { |file| File.exist?(Rails.root.join(file[:url])) }
```
martinlovell: A good place to use all?()

app/models/goobi_xml_import.rb
```diff
@@ -3,15 +3,30 @@
...
   def file=(value)
     @file = value
     self[:goobi_xml] = value.read
   end
 
-  def check_for_oid
+  def oid
     mets_doc = MetsDocument.new(@file)
-    errors.add(:file, 'must contain an oid') if mets_doc.oid.nil? || mets_doc.oid.empty?
+    return errors.add(:file, 'must contain an oid') if mets_doc.oid.nil? || mets_doc.oid.empty?
```
orangewolf, line 15: validation and accessing should be seperate. this method currently has 2 concerns.  I would suggest making def oid return nil and then having a method that validates seperately

spec/fixtures/goobi/metadata/2012315/meta.xml
```diff
@@ -53,34 +53,34 @@
    <mets:fileSec>
       <mets:fileGrp USE="LOCAL">
          <mets:file ID="FILE_0001" MIMETYPE="">
-            <mets:FLocat LOCTYPE="URL" xlink:href="file:///Users/mark/_no_backup_/g2g/goobi/metadata/2012315/images/TEST0003_media/1053442.jpg"/>
+            <mets:FLocat LOCTYPE="URL" xlink:href="file:///spec/fixtures/goobi/metadata/2012315/images/TEST0003_media/1053442.jpg"/>
```
orangewolf: technically (and we're being pretty technical at the moment) file:// is the protocol and /Users/mark was the absolute path.  What you are saying here is that /spec is an absolute path, which it is not and instead these should be `file://spec/fixtures/gooobi/metadata...`. and that means the join you had before was doing the right thing ¯\_(ツ)_/¯ however, since we agreed we are likely changing this again, I think that's probably not a big deal.

orangewolf (changes requested):
one change requested, one informational comment. Good work!

spec/fixtures/goobi/metadata/2012315/meta.xml
maxkadel: AAaaaaaah, I was wondering why there was a triple slash after file, that makes sense.

app/models/mets_document.rb
maxkadel: Can you tell me more? Currently the `all_images_present?` method only returns true if *all* the images referenced in the xml file are present in the file system. I was worried that if I named it "images_present?" it would be confusing if it returned false when 90% of the images were present but not all. Would you think `all_referenced_images_present?` would be clearer?

martinlovell: I meant you could do `files.all? { |file| File.exist?(Rails.root.join(file[:url])) }`
Instead of mapping to the array of booleans and looking for false.

I think all?() will return on the first false and doesn't need to store any booleans in an array.

maxkadel: ah, ok, thanks for the clarification! Sounds like a good improvement, I thought you were referring to the naming of the method, so I'm glad I asked.

orangewolf (approved):

PR #181, opened 2020-07-31
Circle ci wait on release

JzGo:
Adds a script to check for the docker image before the docker pull


orangewolf (approved):

fnibbit:
Testing/troubleshooting this with the existing circleci config requires the publish-github-release circleci workflow to run, which happens only when a numbered (v2.7.0 for instance) release is tagged.  Perhaps test it with a modified config that runs that workflow under other circumstances?

PR #183, opened 2020-08-02
Load samples from S3 if no VPN, cache them to S3 when VPN is on

orangewolf:


spec/spec_helper.rb
```diff
@@ -72,7 +72,8 @@
   #   # is tagged with `:focus`, all examples get run. RSpec also provides
   #   # aliases for `it`, `describe`, and `context` that include `:focus`
   #   # metadata: `fit`, `fdescribe` and `fcontext`, respectively.
-  #   config.filter_run_when_matching :focus
+  config.filter_run_when_matching :focus
+  config.filter_run_excluding vpn_only: true if ENV['VPN'].nil?
```
orangewolf, line 76: this auto filters the vpn only spec based on env. no more forgetting to set the filter =-)

app/lib/activity_stream_reader.rb
```diff
@@ -115,18 +115,13 @@ def refresh_updated_items(parent_objects_for_update)
...
-      full_response = MetadataCloudService.mc_get(metadata_cloud_url)
-      MetadataCloudService.save_mc_json_to_file(full_response, oid, metadata_source)
       po = ParentObject.find_by(oid: oid)
-      if metadata_source == "ladybird"
-        po.last_ladybird_update = DateTime.current
-      elsif metadata_source == "ils"
-        po.last_voyager_update = DateTime.current
-      elsif metadata_source == "aspace"
-        po.last_aspace_update = DateTime.current
+      if po
```
maxkadel, line 119: Non-blocking question: is getting all of the records from the MetadataCloud the behavior we want here in the end? 

I'm having a little trouble knowing how much parallelism we'll want in the end wrt having non-authoritative metadata source records accessible via the Management application. 

I'm thinking as we scale up, perhaps we want to default to only caring about dependent objects if they relate to the authoritative metadata source?

spec/spec_helper.rb
maxkadel: Nice :)

maxkadel (approved):
Looks good! I have a couple questions just out of curiosity / communication, happy to merge as is.

fnibbit (approved):

spec/models/concerns/json_file_spec.rb
```diff
@@ -2,12 +2,12 @@
 require "rails_helper"
 
 RSpec.describe JsonFile do
-  let(:parent_object) { FactoryBot.create(:parent_object) }
-  let(:path_to_example_file) { Rails.root.join("spec", "fixtures", "ladybird", "2004628.json") }
+  let(:parent_object) { FactoryBot.create(:parent_object, oid: '100001') }
```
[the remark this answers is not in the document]
orangewolf: I wanted one that didn’t actually exist. We started with 00000 but they get processed as actual numbers in the json so I needed one that was obviously wrong if it ever go uploaded from the cloud during the test run

app/lib/activity_stream_reader.rb
orangewolf: Great question and one I think we need to pull out of here and discuss with Mark and Mike

PR #187, opened 2020-08-03
Does not try to fetch from MetadataCloud if VPN is set to empty string

maxkadel:
- Setting to empty string seems to be default behavior for docker-compose (not nil)


orangewolf (approved):
I dig it

PR #191, opened 2020-08-03
Will not error out if there is not goobi namespace in xml

maxkadel:
- Pending more information on a more final Goobi xml, give more error info on file upload and don't error out, even if this won't be final form of xml


orangewolf (approved):
This makes sense to me!

PR #193, opened 2020-08-04
Call method in controller to create ParentObject

maxkadel:
An xml import should create a ParentObject, calling the method in the controller that does this.

### DIFF
--- a/spec/system/goobi_xml_imports_spec.rb
+++ b/spec/system/goobi_xml_imports_spec.rb
@@ -5,17 +5,20 @@ RSpec.describe "Goobi Xml Imports", type: :system do
   let(:metadata_cloud_response_body_1) { File.open(File.join(fixture_path, "ladybird", "2012315.json")).read }
 
   before do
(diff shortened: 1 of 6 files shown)


spec/system/goobi_xml_imports_spec.rb
```diff
@@ -5,17 +5,20 @@
   let(:metadata_cloud_response_body_1) { File.open(File.join(fixture_path, "ladybird", "2012315.json")).read }
 
   before do
-    stub_request(:get, "https://metadata-api-test.library.yale.edu/metadatacloud/api/ladybird/oid/2012315")
+    stub_request(:get, "https://yul-development-samples.s3.amazonaws.com/ladybird/2012315.json")
       .to_return(status: 200, body: metadata_cloud_response_body_1)
+    prep_metadata_call
```
orangewolf, line 10: do you need to do this if you are not calling the metadata api?

orangewolf (approved):
Existential question in slack, but that doesn't stop this code from moving forward

spec/system/goobi_xml_imports_spec.rb
maxkadel: It's needed in order to create a parent object

PR #200, opened 2020-08-06
Create manifest from mets

maxkadel:
* Save the oid to the goobi xml import and index on it so the xml can be easily retrieved
* Have the mets document take the raw xml rather than a file, so that it can re-create the mets document and / or manifest as needed


app/lib/iiif_manifest_factory.rb
```diff
@@ -1,22 +1,64 @@
...
-  attr_reader :oid, :manifest, :parent_object
+  attr_reader :oid, :manifest, :parent_object, :mets_doc
 
   require 'iiif/presentation'
 
   def initialize(oid)
     @oid = oid
     @parent_object = ParentObject.find_by(oid: oid)
+    goobi_object = GoobiXmlImport.find_by(oid: oid)
+    return unless goobi_object
```
bess: Seems like if we're requesting a iiif manifest, and not finding something that it needs, we'd want to record that as an error condition, yes? I'd be glad to see that captured in a ticket elsewhere so this could be merged now, but I don't want us creating silent failures.

app/lib/iiif_manifest_factory.rb
```diff
@@ -1,22 +1,64 @@
...
   def construct_manifest
     manifest = IIIF::Presentation::Manifest.new(seed)
+    manifest.sequences << create_first_sequence(oid)
+    add_canvases_to_sequence(manifest.sequences.first)
     manifest
   end
 
+  def create_first_sequence(oid)
+    sequence = IIIF::Presentation::Sequence.new
+    sequence["@id"] = "http://127.0.0.1/manifests/sequence/#{oid}"
```
bess: I believe this is hard-coded as 127.0.0.1 so that it will work with the string substitution elsewhere, right? Would you please add a comment anywhere we're doing that? Otherwise at first glance it looks like we're hard-coding a thing that's wrong. Also, that's the kind of tribal knowledge that gets lost quickly as projects age, so let's drop some pointers for future debuggers of this code. I would also be happy to see a separate ticket that addresses hard-coding the 127.0.0.1 throughout the code base. There might be a better approach here. Again, happy to see this merged as-is as long as we have a plan in place to address it.

bess (changes requested):
This looks great. I love the thorough testing and logical flow. I did flag a couple of places where I'd like to see some changes to help with long-term maintainability, but I'd also be happy to see those captured as tickets so this can be merged as-is.

app/lib/iiif_manifest_factory.rb
orangewolf: I think even if it needs to be set to 127.0.0.1, pulling it out in to its own ENV var is worthy while - it means the "settings" for the app are all in one place and it gives you a chance to give the thing a name. That helps describe the intent and that the group of things are meant to be the same and are not just coincidentally the same.

app/lib/iiif_manifest_factory.rb
orangewolf: I agree, this should throw an exception

app/lib/iiif_manifest_factory.rb
```diff
@@ -1,22 +1,65 @@
...
+  attr_reader :oid, :manifest, :parent_object, :mets_doc
 
   require 'iiif/presentation'
 
   def initialize(oid)
     @oid = oid
     @parent_object = ParentObject.find_by(oid: oid)
+    goobi_object = GoobiXmlImport.find_by(oid: oid)
+    iiif_info = { oid: oid, reason: "No METs xml available for #{oid}" }
+    return Rails.logger.error("Unable to create iiif manifest for: #{iiif_info.to_json}") unless goobi_object
```
orangewolf: @maxkadel I think this needs more Micheal Bay: 💥  I feel like a ActiveRecordNotFound exception is probably appropriate here.

app/lib/iiif_manifest_factory.rb
maxkadel: This should now throw an error, see update.

app/lib/iiif_manifest_factory.rb
maxkadel: See decision document https://docs.google.com/document/d/1TSqKAtitOkgEJbsKJl0NK9FHfd5GP1PRfn8E-8LLak8/edit?usp=sharing 

Sounds like it might make sense to use an environment variable for both manifest host and image host in this PR (aka `IIIF_MANIFESTS_BASE_URL` and `IIIF_IMAGE_BASE_URL`?

orangewolf (approved):
Looking good to me know, thank you!

bess (approved):

PR #202, opened 2020-08-07
Parent object search indexing

orangewolf:
- Added a sidebar nav https://share.getcloudapp.com/d5uEb61n
- Moved indexing in to a concern and assigned it to the parent object to manage
- Removed the FixtureIndexingService and all calls to it (its the very last commit, just in case we decide we dont want that after all)
- Added a spec helper for solr specs (tag :solr) and added a helper for request stubs


maxkadel (approved):
I really like these changes, especially the spec helpers, the hook with the parent object creation for the indexing to solr, and adding the sidebar. I think those all add a ton of value.

Two things we might want to update on a future iteration: 
* There's currently a "clean" spec configuration that cleans out Solr, I think your solr helper method is better but we should remove the other so we don't have duplication. (see https://github.com/yalelibrary/yul-dc-management/blob/362fd81c631b7c69999f523548a19b1ff13fbe7e/spec/rails_helper.rb#L47)
* We should update the ReadMe to reflect the change in the rake task

PR #206, opened 2020-08-10
Refactor tests to make more readable

maxkadel:
- Remove superceded spec helper
- Use new stub helper to make tests more readable


fnibbit (approved):

orangewolf (approved):
Thank you for doing all the leg work here!

PR #209, opened 2020-08-12
Child objects

maxkadel:
Connected to https://github.com/yalelibrary/YUL-DC/issues/418


app/models/parent_object.rb
```diff
@@ -6,6 +6,7 @@ class ParentObject < ApplicationRecord
   include JsonFile
   include SolrIndexable
   has_many :dependent_objects
+  has_many :child_objects
```
orangewolf: Defendant destroy?

orangewolf (approved):
One tiny thing - we can pick it up now or grab it later.

PR #219, opened 2020-08-18
Image conversion lib & test

maxkadel:
* includes libvips-tools & imagemagick & libtiff-tools


lib/nga.sh
```diff
@@ -0,0 +1,134 @@
+#!/bin/bash
```
maxkadel: More descriptive file name? It's a bit cryptic.

lib/nga.sh
```diff
@@ -0,0 +1,134 @@
...
+    if [ -z "${savefiles}" ]  && [ -d $tmpprefix ]; then rm -f $outprefix*; fi
+}
+
+trap 'handle_error ${LINENO} $?' ERR
+
+# cause the script to fail if any commands exist with non-zero status
+set -e
+
+if [[ -z "$1" || -z "$2" || -z "$3" ]]; then
+    echo "Usage:tiff_to_pyramid.bash <tmpdir> <full path to image source> <full path to output target> <save all working files - any value will do>";
```
fnibbit: Usage message doesn't match script name.

lib/nga.sh
```diff
@@ -0,0 +1,134 @@
...
+
+if [ ! -f $3 ]; then
+    echo "Could not create the output file.";
+    exit 1;
+fi
+
+rm -f $3
+
+tmpdir=$1
+# only process the first page of a potentially multi-page image document
```
fnibbit: What does this comment apply to?

lib/nga.sh
```diff
@@ -0,0 +1,134 @@
...
+
+# first, use vipsheader to read the bands
+CHANNELS=$(identify -format "%[channels]\n" ${input}[0] 2>/dev/null)
+echo "channels: ${CHANNELS}"
+if [ ${CHANNELS} = "srgba" ]; then
+    # we have to flatten the image to remove the alpha channel / trasparency before proceeding
+    echo "removing alpha channel from $input"
+    vips im_extract_bands $input ${input}.noalpha.tif 0 3   2>&1
+    if [ -z "${savefiles}" ]; then rm $input; fi
+    mv ${input}.noalpha.tif $input
```
fnibbit: Looks like the original file gets overwritten whether or not ${savefiles} is set -- am I misreading this?  I see that ${savefiles} is unset when we use this tool from lib/yale_convert.rb though.

lib/nga.sh
mlooney: I have no idea. this is basically a copy of the NGA's script, with minor adjustments to paths & such

lib/nga.sh
mlooney: yeah, lots of fun stuff in this script. it'll also delete itself if temp is an empty string...

lib/nga.sh
mark-dce: It feels like this and the next comment can be resolved by renaming the script `tiff_to_pyramid.bash`.  Let's do that.

lib/nga.sh
mark-dce: It feels like this and the previous comment can be resolved by renaming the script `tiff_to_pyramid.bash`.

lib/nga.sh
mark-dce: Mike was very clear about wanting to run the NGA scripts unmodified.   Let's ensure that the rest of our code encapsulates them cleanly, but let's not attempt to modify their internals (even opaque comments) and redirect any questions about how the code functions back to whoever made the scripts at the NGA

lib/nga.sh
mark-dce: Let's make sure we cleanly encapsulate these scripts when calling them, but try not to modify them from the NGA originals.

mark-dce (changes requested):
Let's rename the script to `tiff_to_pyramid.bash`.  Beyond that, this looks like a good place to start, so let's get it merged.

fnibbit (approved):

mark-dce (approved):

PR #223, opened 2020-08-20
Refactor manifest creation to be from DB rather than from METS document

maxkadel:
Connected to https://github.com/yalelibrary/YUL-DC/issues/467


fnibbit (approved):
Makes sense to me -- does the "with_index" thing in ParentObject create_child_records control the ChildObjects coming back in order in IiifManifestFactory add_image_to_canvas?

maxkadel:
> Makes sense to me -- does the "with_index" thing in ParentObject create_child_records control the ChildObjects coming back in order in IiifManifestFactory add_image_to_canvas?

Ah, yeah, not so much. Fixed.

orangewolf (approved):

PR #229, opened 2020-08-26
WIP: 468 IIIF Presentation Validations

dylansalay:


app/models/iiif_presentation.rb
```diff
@@ -1,39 +1,45 @@
...
     @manifest_base_url = ENV["IIIF_MANIFESTS_BASE_URL"] || "http://localhost/manifests/"
     @image_base_url = ENV["IIIF_IMAGE_BASE_URL"] || "http://localhost:8182/iiif"
-    @oid = oid
     @parent_object = ParentObject.find_by(oid: oid)
-    @manifest = construct_manifest
+    @oid = oid
+  end
+
+  def validate_images
+    raise "There are no child objects for #{@oid}" if ChildObject.where(parent_object: parent_object).order(:order).empty?
```
orangewolf, line 16: we shouldn't just `raise "string"` It would be better to wrap this in its own exception class so it could be caught and handled.  Having said that it would be much better if this never threw an exception at all.

orangewolf, line 16: This would be a better approach:

```suggestion
    if ChildObject.where(parent_object: parent_object).order(:order).empty?
      errors.add(:manifest, 'There are no child objects for #{@oid}') 
      return false
    end
    true
```

Then you don't have to recover from a raise

app/models/iiif_presentation.rb
```diff
@@ -1,39 +1,45 @@
...
 
   require 'iiif/presentation'
 
   def initialize(oid)
     @manifest_base_url = ENV["IIIF_MANIFESTS_BASE_URL"] || "http://localhost/manifests/"
     @image_base_url = ENV["IIIF_IMAGE_BASE_URL"] || "http://localhost:8182/iiif"
-    @oid = oid
     @parent_object = ParentObject.find_by(oid: oid)
-    @manifest = construct_manifest
+    @oid = oid
```
orangewolf, line 12: see https://api.rubyonrails.org/classes/ActiveModel/Errors.html

orangewolf:
@dylansalay @jpengst 2 things - there is no valid? method any more which doesn't make sense to me. we're calling it in the parent_object code, so I think it needs to exist. 

This code raises an exception or returns nil in several places which are going to cause problems with being able to actually use them.  They aren't "in unforseen state" if they don't have children and we don't want them to fail silently and carry on either.

orangewolf (changes requested):

bess:
@orangewolf FYI, I'm going to go ahead and cherry-pick your "Refactor iiif_manifest_factory in to a PORO pattern" commit into `master` b/c Max and I want to build on top of it and it and don't want to create a conflict tangle.

bess:
Welp, I hope I didn't just make things more confusing. Profuse apologies if so. From the history on this PR I thought there was a lot more happening here and I just wanted to get Rob's refactor over the the finish line. I've done that, here: https://github.com/yalelibrary/yul-dc-management/pull/232

But now, I see that there were some suggestions that @orangewolf had made that didn't get incorporated into that PR. Given that the history on this branch is difficult to follow, could we start with a fresh pull off master, which will now include the bulk of this refactor, and pull in the suggested changes on it?

PR #231, opened 2020-08-26
479 Add author_tesim and subject_topic_tesim fields

dylansalay:
This PR adds author_tesim and subject_topic_tesim fields to allow search on keywords within these fields from the search bar.
I've confirmed that both fields are showing up in Solr.

### DIFF
--- a/app/models/concerns/solr_indexable.rb
+++ b/app/models/concerns/solr_indexable.rb
@@ -36,7 +36,6 @@ module SolrIndexable
       alternativeTitleDisplay_tesim: json_to_index["alternativeTitleDisplay"],
       archiveSpaceUri_ssi: json_to_index["archiveSpaceUri"],
       author_ssim: json_to_index["creator"],
-      author_tsim: json_to_index["creator"],
       author_tesim: json_to_index["creator"],
       box_ssim: extract_box_ssim(json_to_index),
       caption_tesim: json_to_index["caption"],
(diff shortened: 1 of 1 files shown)


app/models/concerns/solr_indexable.rb
```diff
@@ -129,6 +130,7 @@ def to_solr(json_to_index = nil)
       sourceNote_ssim: json_to_index["sourceNote"], # replaced by sourceNote_tesim
       sourceTitle_ssim: json_to_index["sourceTitle"], # repleaced by sourceTitle_tesim
       subject_topic_tsim: json_to_index["subjectTopic"], # replaced by subjectTopic_tesim and subjectTopic_ssim
+      subject_topic_tesim: json_to_index["subjectTopic"], # replaced by subjectTopic_tesim and subjectTopic_ssim
```
maxkadel: This section is for things that will be removed in future, so it doesn't make sense to me to add to it. 

Previously this method used inconsistent naming for the Solr fields - either ruby-esque, like `subject_topic_tesim` or not-so-ruby, `subjectTopic_tesim`. Since the majority of the Solr fields, already present in Blacklight and other Samvera projects, use the camelCase naming for Solr indexing, I thought it made more sense to move the few exceptions to that convention. I would not add `subject_topic_tesim` to this section, but instead would communicate to the Blacklight folks that the field they need to look for is `subjectTopic_tesim`. 

I *would* add `author_tsim` to this section and add a comment that it's been replaced by `author_tesim`.

maxkadel (changes requested):

app/models/concerns/solr_indexable.rb
dylansalay: @maxkadel  Updated

maxkadel: But move `author_tsim` to the bottom section (the section for things that will be deprecated in a future release) and add a note that it's been replaced by `author_tesim`.

While it's not bad to have a certain amount of duplication in the Solr index, we want to avoid it accumulating too much cruft.

maxkadel: Nevermind, think you were fixing this as I typed. Or I was viewing an older version. Sorry.

maxkadel (approved):

PR #233, opened 2020-08-27
Refactor valid? method and updated specs

jpengst:
Added validation method to check errors on parent object if there are no images.


orangewolf (approved):

spec/models/iiif_presentation_spec.rb
```diff
@@ -138,4 +138,70 @@ def stub_info
...
+      canvas.width = 191
+      expect { canvas.to_json }.to raise_error('A(n) height is required for each IIIF::Presentation::Canvas')
+    end
+
+    it 'raises an exception if the images service_id is not included' do
+      expect { IIIF::Presentation::ImageResource.create_image_api_image_resource }.to raise_error('key not found: :service_id')
+    end
+
+    # #TODO: Fix this test. Not currently working as at should since we are stubbing requests
+    xit 'raises an exception if the images service_id is not valid' do
```
maxkadel, line 180: Do you want to get this passing before merging? I could help you with how to stub it, I think.

PR #235, opened 2020-08-28
Refactor child objects to use "oid"

maxkadel:
- rather than "child oid" as identifier
- leads to more re-usable code


bess (approved):
Good idea to refactor this. :)

PR #245, opened 2020-09-02
Update metadata when record is saved

orangewolf:
When metadata authroity is set or when metadata_update is true, update the metadata and reindex the object. Fixes https://app.zenhub.com/workspaces/yul-dc-5e50083561915b11fc9e5850/issues/yalelibrary/yul-dc/520


maxkadel (approved):
Looks good - added a test.

PR #248, opened 2020-09-03
I499 index captions and labels

dylansalay:
This PR adds child captions and labels as searchable multivalued fields on the parent object. 

Screen Shot 2020-09-03 at 10 02 10 AM


maxkadel:
Superseded

PR #249, opened 2020-09-03
I525 index oids for childobjects

dylansalay:
This PR adds child oids as searchable multivalued fields on the parent object.

Screen Shot 2020-09-08 at 10 56 13 AM


.rubocop_todo.yml
```diff
@@ -9,7 +9,7 @@
 # Offense count: 2
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 114
+  Max: 140
```
orangewolf: This is global and because it’s the TODO file it will get accidentally deleted or removed. Instead disable a class locally with #rubocop/disable Metrics/ClassLength

app/models/parent_object.rb
```diff
@@ -139,4 +139,28 @@ def iiif_manifest
     return @iiif_manifest if @iiif_manifest
     @iiif_manifest = @iiif_presentation.manifest if iiif_presentation.valid?
   end
+
+  def child_captions
+    captions = []
```
orangewolf: This implementation is correct, but not optimal. Let’s spend a couple minutes this morning tidying this up

orangewolf (changes requested):

app/models/parent_object.rb
dylansalay: This has been refactored

orangewolf (approved):

PR #253, opened 2020-09-10
I542 csv with multiple sources

dylansalay:
Allows CSV to declare the source in a column with header "Source" (ladybird, ils, aspace) and sets the authoritative metadata source when the OID is ingested. The default authoritative metadata source is ladybird.

### DIFF
--- a/spec/services/metadata_cloud_service_spec.rb
+++ b/spec/services/metadata_cloud_service_spec.rb
@@ -29,7 +29,7 @@ RSpec.describe MetadataCloudService, prep_metadata_sources: true do
     end
     it "can create a parent_object from an array of oids" do
       expect(ParentObject.count).to eq 0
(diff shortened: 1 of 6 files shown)


spec/services/metadata_cloud_service_spec.rb
```diff
@@ -92,7 +92,7 @@
       time_stamp_before = File.mtime(path_to_example_file.to_s)
       described_class.refresh_fixture_data(short_oid_path, metadata_source)
       time_stamp_after = File.mtime(path_to_example_file.to_s)
-      expect(time_stamp_before).to be < time_stamp_after
+      expect(time_stamp_before).to be <= time_stamp_after
```
dylansalay: time_stamp before and after are the same after our changes. Do we need to look into this further?

orangewolf (approved):

PR #269, opened 2020-09-15
WIP: Add notification reporting for CSV imports

dylansalay:


spec/requests/notifications_request_spec.rb
```diff
@@ -0,0 +1,12 @@
...
+require 'rails_helper'
+
+RSpec.describe "Notifications", type: :request do
+  describe "GET /index" do
+    it "returns http success" do
+      get '/notifications'
+      expect(response).to redirect_to(user_session_path)
+    end
+  end
+end
```
dylansalay, line 49: I think this needs to be fixed. We are authenticating_user for Notification controller actions but I'm not sure how to simulate logging in within a request spec (devise login_user helpers don't work in request specs). I will dig into this

maxkadel: I wonder if you can create a user with FactoryBot and then use login_as(user)

```
let(:user) { FactoryBot.create(:user) }

it "can get to notifications when logged in" do
login_as(user)
get .....
end
it "gets redirected to login when not logged in" do
get....
end
```

maxkadel: Oh, someone already did this in the latest. Cool!

app/views/notifications/index.html.erb
```diff
@@ -0,0 +1,44 @@
+<p id="notice"><%= notice %></p>
+<div class='row'>
+  <div class='col'>
+    <h1>Notifications</h1>
+  </div>
+  <div class='col'>
+    <div class='float-right'>
+      <%#= link_to 'Reindex', reindex_parent_objects_path, method: 'post', class: 'btn btn-primary' %>
```
maxkadel: Remove these commented lines?

spec/models/metadata_source_spec.rb
```diff
@@ -38,6 +42,12 @@
         expect(ladybird_result).to be
         expect(ladybird_result['uri']).to eq('/ladybird/oid/16797069')
       end
+
+      it 'adds a notificaiton if record is not found in metadata cloud' do
```
maxkadel, line 46: I like this test!

maxkadel (approved):
Looks good to merge! May want to remove stray code in `app/views/notifications/index.html.erb` before doing so, but ok to merge without.

app/views/notifications/index.html.erb
dylansalay: @orangewolf are there some links you wanted me to add here? Or can we delete these for now

spec/views/notifications/index.html.erb_spec.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe "notifications/index.html.erb", type: :view do
+  pending "add some examples to (or delete) #{__FILE__}"
+end
```
dylansalay, line 7: Do we need this file?

maxkadel (approved):

spec/views/notifications/index.html.erb_spec.rb
maxkadel: I could see us going either way. We have a lot of pending tests that we haven't discussed. I think keeping the placeholder, and then having a "how do we want to deal with system specs vs. view specs" larger conversation, especially since it sounds like we're about to shift more towards UI stuff.

app/views/notifications/index.html.erb
orangewolf: remove is fine for now

PR #282, opened 2020-09-18
Test putting shared gems in base docker build

maxkadel:


maxkadel:
Did not speed docker build, closing.

PR #293, opened 2020-09-22
Parent object datatable

orangewolf:


app/datatables/parent_object_datatable.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+class ParentObjectDatatable < AjaxDatatablesRails::ActiveRecord
+  def view_columns
+    # Declare strings in this format: ModelName.column_name
+    # or in aliased_join_table.column_name format
+    @view_columns ||= {
```
maxkadel, line 7: Future enhancement: Change the condition to match the field types more closely (numeric ids should probably be `:eq`)

maxkadel (changes requested):
Looks good! Found a bug, though -
When I first click on the sidebar for "Parent Objects," the table only shows the headers. When I refresh the page, it shows the parent objects. Even after having done this, when I click away and click back again, I'm getting the same behavior - empty table, then refresh and can see the parent objects.

app/datatables/parent_object_datatable.rb
orangewolf: I played with this quite a bit - the 'like' is really nice on these because they filter down as you type it in, but it may not scale when we load a lot of records.

orangewolf:
TURBOLINKS! (I know how to fix that)

maxkadel (approved):

PR #296, opened 2020-09-23
visibility dropdown private public yale community only

MaggieZhaoYale:
https://app.zenhub.com/workspaces/yul-dc-5e50083561915b11fc9e5850/issues/yalelibrary/yul-dc/607
Added: 

- 3 levels visibilities:  'Private', 'Public', 'Yale Community Only' in visibility dropdown (phrase1)

- validates :visibility in model of parent_object

Screen Shot 2020-09-23 at 2 04 48 PM

### DIFF
--- a/app/models/parent_object.rb
+++ b/app/models/parent_object.rb
@@ -9,6 +9,8 @@ class ParentObject < ApplicationRecord # rubocop:disable Metrics/ClassLength
   has_many :child_objects, primary_key: 'oid', foreign_key: 'parent_object_oid', dependent: :destroy
   belongs_to :authoritative_metadata_source, class_name: "MetadataSource"
   attr_accessor :metadata_update
+  validates :visibility, inclusion: { in: ['Private', 'Public', 'Yale Community Only'],
+                                message: "%{value} is not a valid value" }
 
   self.primary_key = 'oid'
   after_save :setup_metadata_job
--- a/app/models/parent_object.rb
+++ b/app/models/parent_object.rb
@@ -9,14 +9,18 @@ class ParentObject < ApplicationRecord # rubocop:disable Metrics/ClassLength
   has_many :child_objects, primary_key: 'oid', foreign_key: 'parent_object_oid', dependent: :destroy
   belongs_to :authoritative_metadata_source, class_name: "MetadataSource"
   attr_accessor :metadata_update
-  validates :visibility, inclusion: { in: ['Private', 'Public', 'Yale Community Only'],
-                                      message: "%{value} is not a valid value" }
-
   self.primary_key = 'oid'
   after_save :setup_metadata_job
   after_update :solr_index_job # we index from the fetch job on create
   after_destroy :solr_delete
   paginates_per 50
+
+  def self.visibilities
+    ['Private', 'Public', 'Yale Community Only']
+  end
+
+  validates :visibility, inclusion: { in: visibilities,
+                                      message: "%{value} is not a valid value" }
 
   def create_child_records
     return unless ladybird_json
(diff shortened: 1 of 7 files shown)


app/models/parent_object.rb
```diff
@@ -9,6 +9,8 @@ class ParentObject < ApplicationRecord # rubocop:disable Metrics/ClassLength
   has_many :child_objects, primary_key: 'oid', foreign_key: 'parent_object_oid', dependent: :destroy
   belongs_to :authoritative_metadata_source, class_name: "MetadataSource"
   attr_accessor :metadata_update
+  validates :visibility, inclusion: { in: ['Private', 'Public', 'Yale Community Only'],
+                                      message: "%{value} is not a valid value" }
```
orangewolf: please add a class method called visibilities like so:

```
def self.visibilities
  ['Private', 'Public', 'Yale Community Only']
end
```
then use that class method instead of the hard coded string here and in the view. That way, if we ever have to add one, we can easily

app/views/parent_objects/_form.html.erb
```diff
@@ -63,7 +63,7 @@
 
   <div class="field">
     <%= form.label :visibility %>
-    <%= form.text_field :visibility %>
+    <%= form. select(:visibility,[['Private','Private'],['Public','Public'],['Yale Community Only','Yale Community Only']] ) %>
```
orangewolf: I think you can declair these once `['Private', 'Public', 'Yale Community Only']` and Rails will automatically do the doubling. I could be wrong about that... if I am, then `ParentObject.visibilities.map {|v| [v,v] }` would work

orangewolf (dismissed):
Looks good on the whole, I think we have an opportunity to reduce some duplication.  I'd also like to see a test of the model validation please. I personally think its fine to skip the view spec for this kind of small change, but others may want to weigh in there.

app/models/parent_object.rb
MaggieZhaoYale: @orangewolf Thank you, that is a good suggestion. I will add a class method.

maxkadel: > Looks good on the whole, I think we have an opportunity to reduce some duplication. I'd also like to see a test of the model validation please. I personally think its fine to skip the view spec for this kind of small change, but others may want to weigh in there.

The general DCE pattern is to use system specs for these types of changes, rather than view specs (I don't have a lot of experience with view specs, but they seem a little implementation-focused to me, which I think is a drawback). I do think adding this to the ParentObject system spec would be valuable in this case.

maxkadel (changes requested):
Agree with Rob's requests, would like the dropdown added to the system spec for ParentObjects.

MaggieZhaoYale:
@maxkadel @orangewolf Thank you. Will commit the change later today.

MaggieZhaoYale:
spec/model/parent_object_spec.rb 208 the order of child_oid is different from the expected. So I slightly modified it.
Screen Shot 2020-09-24 at 1 03 26 PM

spec/models/parent_object_spec.rb
```diff
@@ -235,4 +235,15 @@
       end
     end
   end
+
+  context "with correct visibility value" do
```
maxkadel, line 259: We need to test for what happens for an empty visibility from the metadata cloud, and for a visibility other than the 3 we've defined in the model is given. Basically, should test the validation method for both nil and random strings.

Also, @orangewolf , should we make the visibilities a private method?

maxkadel (changes requested):

maxkadel (approved):

maxkadel (approved):

PR #298, opened 2020-09-23
Require CAS authentication for all pages 

bess:
* Protect all routes in the management application behind CAS authentication.
* Update tests to match this expectation.

Connected to https://github.com/yalelibrary/YUL-DC/issues/602

Co-authored-by: @maxkadel 
Co-authored-by: @orangewolf

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -37,6 +37,4 @@ Rails.application.routes.draw do
   authenticated :user do
     mount DelayedJobWeb, at: "/delayed_job"
   end
-  # fall back if not authenticated
-  # get '/delayed_job', to: redirect(user_cas_omniauth_authorize_path)
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -37,4 +37,6 @@ Rails.application.routes.draw do
   authenticated :user do
     mount DelayedJobWeb, at: "/delayed_job"
   end
+  # fall back if not authenticated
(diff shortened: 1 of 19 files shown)


config/routes.rb
```diff
@@ -37,6 +37,4 @@
   authenticated :user do
     mount DelayedJobWeb, at: "/delayed_job"
   end
-  # fall back if not authenticated
-  get '/delayed_job', to: redirect('/management/users/sign_in')
```
orangewolf, line 41: I am pretty sure this should be `get '/delayed_job', to: redirect('/management/users/auth/cas')`

orangewolf (dismissed):
Small change to fix in delayed_job login exception

config/routes.rb
bess: Fixed

bess:
@orangewolf Is this what you meant when you said we should raise an exception to prevent a redirect loop? 

```
    def after_omniauth_failure_path_for(_resource)
      raise 'CAS Authentication is not working'
    end

```

maxkadel (approved):

bess:
When we tried to deploy this branch we realized we need a place outside of CAS that the health checks can look at to ensure the application is running. After conversation with @mark-dce we decided to expose part of the management#index page, just enough that it prompts you to log in if you go to the root path of the application without authentication: 

Screen Shot 2020-09-24 at 11 27 18 AM

maxkadel (approved):

PR #299, opened 2020-09-24
Remove expectation of User.email (#297)

bess:
The User email field is not populated by CAS, so remove it from the User 
FactoryBot factory, and remove the database index for that field.


maxkadel:
Already in, was put in to fix a release oddity

PR #301, opened 2020-09-25
Change default expectations for parent objects based on conversations

maxkadel:
- Basically, we don't want to make assumptions about our parent objects, and only want to record things in the model that we know for sure. 
- Leaving the assumption of Ladybird right now, because I think we'll need a larger pass & more information to change that assumption


MaggieZhaoYale (approved):

MaggieZhaoYale:
Goods look, and ran the rspec test, all passed. There are some conflicting in the spec.

orangewolf (approved):

maxkadel:
This has gone stale, will re-open more built out

PR #302, opened 2020-09-25
I608 clickable oids

jpengst:
User can click on the oid in the Parent Objects datatable to route to that Parent Object showpage.


maxkadel (dismissed):
I don't think this will be merge-able until we get the asset pipeline working again in production.

martinlovell (approved):
Looks good

martinlovell:
Merge conflicts.

PR #308, opened 2020-09-29
add discover button

MaggieZhaoYale:
To fix #https://app.zenhub.com/workspaces/yul-dc-5e50083561915b11fc9e5850/issues/yalelibrary/yul-dc/610

Screen Shot 2020-09-29 at 3 42 51 PM

### DIFF
--- a/app/controllers/parent_objects_controller.rb
+++ b/app/controllers/parent_objects_controller.rb
@@ -90,6 +90,11 @@ class ParentObjectsController < ApplicationController
     end
   end
 
+  def blacklight_url(path)
+    base = ENV['BLACKLIGHT_BASE_URL'] || 'localhost:3000'
+    "#{base}/catalog/#{path}"
+  end
+
   private
 
     # Use callbacks to share common setup or constraints between actions.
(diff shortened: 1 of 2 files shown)


dylansalay (approved):

martinlovell (approved):

martinlovell:
The data tables were merged in, so this will have to change to use data tables.

maxkadel (changes requested):
This looks good - unfortunately it doesn't align with the datatables work that was just merged (sorry that's gone back and forth a couple times). 

I'm not sure whether trying to resolve the conflicts would be easier, or pulling from master and re-doing the view part of the work from there.

Also, whenever we add a new feature / expectation to the application, let's try to also write a test / tests for it. 

So, for this, we could write a spec that sets up an environment variable for BLACKLIGHT_BASE_URL, and then expect that a call to the helper method returns the expected string. We could also add it to the system spec - expect the button, expect that clicking the button directs you to a given url.

MaggieZhaoYale:
@maxkadel I will redo it. It looks that there are a lot of change after the merging of datatable.

MaggieZhaoYale:
close this one. Redo after pull the change of datatable

PR #312, opened 2020-10-01
Improve indexing performance overall and especially in development mode

orangewolf:
- Reorder the jobs so that a) we prefer to get all metadata in as fast as possible (so records can be checked) and then prefer to do ptiff -> manifest -> index a single record instead of trying to do all ptiff jobs before building a single manifest (which is what currently happens on an import).  This part applies globally

- Look for cached child widths and heights and if they exist, then use those instead of creating the PTIFF. Right now there is fall back width and height but that is only because it was taking to long to make them on my machine. This can be replaced once a larger set it created and saved to JSON. This change currently only applies to development. It means loading the 45 sample objects in less than 2 minutes on my machine


FCRodriguez7 (changes requested):
Over the "Parent Object" when one clicks on the button "reindex" button the application provides an error. 

Screen Shot 2020-10-01 at 8 43 08 AM

FCRodriguez7:
There's something wrong with my yale credential when I use my DCE credential to access the bucket it works.

FCRodriguez7 (approved):

PR #313, opened 2020-10-01
discover in the datatable data

MaggieZhaoYale:
move the blacklight_url from controller to datatable/parent_object_datatable.rb. 
add the button in the datatable link
add spec for the discover link/button in the parent_object_datatable_spec.rb


maxkadel (changes requested):
Looking good!

It sounds like we want to wait on a confirmation of wording before we merge this ("Discover" or "Discovery" or something else), but once we have the correct wording for the button I think we can merge this.

MaggieZhaoYale:
@maxkadel Just realized that I need to put the ENV['BLACKLIGHT_BASE_URL'] || 'localhost:3000' instead of localhost:3000 in the test.
I will have the new commit.

maxkadel (approved):

PR #314, opened 2020-10-01
Fix datatable route

maxkadel:
- I think we need to work on the testing for this feature (system spec, etc.)


MaggieZhaoYale (approved):
Tested and  looks great

PR #329, opened 2020-10-08
Add extent of digitization and rights statement to the parent object …

dl-maura:
…model and views etc

### DIFF
--- a/app/models/parent_object.rb
+++ b/app/models/parent_object.rb
@@ -22,6 +22,13 @@ class ParentObject < ApplicationRecord # rubocop:disable Metrics/ClassLength
   validates :visibility, inclusion: { in: visibilities,
                                       message: "%{value} is not a valid value" }
 
+  def self.digi_extents
+    ['Complete folder digitized.', 'Complete work digitized.', 'Completely digitized.', 'Partial work digitized.', 'Partial folder digitized.', 'Partial collection digitized.', 'Partially digitized.']
+  end
+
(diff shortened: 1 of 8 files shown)


app/models/parent_object.rb
```diff
@@ -98,6 +105,8 @@ def ladybird_json=(lb_record)
     self.barcode = lb_record["orbisBarcode"]
     self.aspace_uri = lb_record["archiveSpaceUri"]
     self.visibility = lb_record["itemPermission"]
+    self.extent_digitization = lb_record["extentOfDigitization"][0]
```
maxkadel, line 108: If we're getting this out of an array, I think it's more ruby-ish to use `.first` instead of the index. Same with next row. (this is just a style preference thing, not right / wrong)

app/views/parent_objects/_form.html.erb
```diff
@@ -66,6 +66,16 @@
...
   </div>
 
+  <div class="field">
+    <%= form.label :extent_digitization %>
+    <%= form.select(:extent_digitization, ParentObject.digi_extents) %>
+  </div>
+
+  <div class="field">
+    <%= form.label :rights_statement %>
+    <%= form.text_field :rights_statement %>
```
maxkadel, line 76: Curious - will we eventually want this to be a select list as well?

maxkadel (changes requested):
Looking forward to checking out the tests!

maxkadel:
Closing for now, can refer back to this as needed, but it's getting a bit stale.

PR #330, opened 2020-10-08
Update deploy instructions

alishaevn:
# Summary
add instructions on how to deploy to the yale infrastructure

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,26 +206,46 @@ In production and development mode, you can add objects either individually, eit
 Any time you pull a branch with a Gemfile change you need to pull or build a new Docker image. If you change the Dockerfile, you need to build a new Docker image. If you change a file in ./ops you need to build a new Docker image. These are the primary times in which you need to pull or build.
 
 ## Releasing a new version
+1. Checkout to the `master` branch and run `git pull`
 
-1. Ensure you have a github personal access token.
+2. Ensure you have a github personal access token.
 
-  Instructions here: <https://github.com/github-changelog-generator/github-changelog-generator#github-token> You will need to make your token available via an environment variable called `CHANGELOG_GITHUB_TOKEN`, e.g.:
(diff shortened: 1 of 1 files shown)


maxkadel (changes requested):
This refers to the blacklight version several times, rather than management. 

I'm wondering if this would more appropriately live in the Camerata repo, since it has more to do with all the projects, and only include things that are specific to the management app here?

alishaevn:
> This refers to the blacklight version several times, rather than management.
> 
> I'm wondering if this would more appropriately live in the Camerata repo, since it has more to do with all the projects, and only include things that are specific to the management app here?

whoops! thought I had caught those. 


mmm. I guess that's a discussion to be had. my preference though is having the instructions in the repo I'm working with instead of needing to bounce between two.

alishaevn:
removed the references to blacklight

maxkadel:
> > This refers to the blacklight version several times, rather than management.
> > I'm wondering if this would more appropriately live in the Camerata repo, since it has more to do with all the projects, and only include things that are specific to the management app here?
> 
> whoops! thought I had caught those.
> 
> mmm. I guess that's a discussion to be had. my preference though is having the instructions in the repo I'm working with instead of needing to bounce between two.

That makes sense - my concern is that documentation goes stale so quickly, and if we try to maintain the same information on each repo, it's going to get out of sync really quickly. I'd rather have documentation that will be the same for every repo in the project be maintained in a single central place, so that it only has to be maintained there.

alishaevn:
> That makes sense - my concern is that documentation goes stale so quickly, and if we try to maintain the same information on each repo, it's going to get out of sync really quickly. I'd rather have documentation that will be the same for every repo in the project be maintained in a single central place, so that it only has to be maintained there.

@maxkadel  valid point if there were all the same. 
releasing cam doesn't appear to be the same as blacklight/management though. it doesn't have anything about the github token in its readme and the yale infrastructure release doesn't apply either. so I think it would wind up being a bunch of "if / else if / else" statements in the camerata repo that could also lead the user back to the repo they're actually deploying for if it's something specific to that repo.

maxkadel (approved):

PR #331, opened 2020-10-08
Batch process event

maxkadel:
Connected to https://github.com/yalelibrary/YUL-DC/issues/647


maxkadel:
Going a different direction

PR #335, opened 2020-10-12
Create a manifest even if PTIFFs already exist in environment

maxkadel:


app/models/pyramidal_tiff.rb
```diff
@@ -27,12 +27,10 @@ def generate_ptiff
   end
 
   def verify_and_generate
-    ptiff_info = { oid: oid.to_s }
     # cannot convert to PTIFF if we can't find the original
     return false unless original_file_exists?
     # do not do the image conversion if there is already a PTIFF on S3
     if child_object.height && child_object.width && S3Service.s3_exists?(child_object.remote_ptiff_path)
-      errors.add(:base, "PTIFF exists on S3, not converting: #{ptiff_info.to_json}")
```
orangewolf, line 35: I think there should be a notification (new kind) at this spot, I was kind of leaving the error in until we got that.  I'd be fine with just a big old "TODO add event here" since that is close now.

orangewolf (changes requested):

orangewolf (approved):

orangewolf:
This needs a rename @maxkadel but can then be merged (just the PR name)

PR #345, opened 2020-10-15
Spit out iam creds during seeeds, for debugging roles

fnibbit:


orangewolf:
we're done with this test code. we thank it for its service and wish it well.

PR #346, opened 2020-10-16
Allows unauthenticated access to oid minter

martinlovell:


mikeapp:
Can you change the test as well, there's a login in a few places:
```
      before do
        login_as(user)
      end
```

orangewolf (approved):

PR #350, opened 2020-10-16
Ability to still show import history if parent object is deleted

orangewolf:
- Batch Process show page
- Notifications Page

### DIFF
--- a/app/views/batch_processes/show.html.erb
+++ b/app/views/batch_processes/show.html.erb
@@ -25,7 +25,7 @@
     <% @batch_process.oids.each do |parent_oid| %>
       <tr>
         <td class="parent_oid"><%= link_to parent_oid, parent_object_path(parent_oid) %></td>
-        <td class="child_count"><%= ParentObject.find(parent_oid).child_object_count.to_s %></td>
+        <td class="child_count"><%= ParentObject.find_by_oid(parent_oid)&.child_object_count || "(deleted)" %></td>
       </tr>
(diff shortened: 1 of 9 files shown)


db/schema.rb
```diff
@@ -87,6 +87,14 @@
     t.index ["parent_object_id"], name: "index_dependent_objects_on_parent_object_id"
   end
 
+  create_table "metadata_samples", force: :cascade do |t|
```
maxkadel: These need to be re-removed (your local database re-created them)

app/views/batch_processes/show.html.erb
```diff
@@ -25,7 +25,7 @@
     <% @batch_process.oids.each do |parent_oid| %>
       <tr>
         <td class="parent_oid"><%= link_to parent_oid, parent_object_path(parent_oid) %></td>
-        <td class="child_count"><%= ParentObject.find(parent_oid).child_object_count.to_s %></td>
+        <td class="child_count"><%= ParentObject.find_by_oid(parent_oid)&.child_object_count || "(deleted)" %></td>
```
maxkadel, line 28: Why #find_by_oid when presumably #find also works? (since the oid is the ID for the object?) 

I'm going to check this out to see whether I can get it passing CI, so I'll double check that I'm right about that assumption.

maxkadel (changes requested):

maxkadel (approved):

PR #351, opened 2020-10-19
Disable link if parent object doesn't exist

maxkadel:


app/views/batch_processes/show.html.erb
```diff
@@ -24,7 +24,7 @@
   </thead>
     <% @batch_process.oids.each do |parent_oid| %>
       <tr>
-        <td class="parent_oid"><%= link_to parent_oid, parent_object_path(parent_oid) %></td>
+        <td class="parent_oid" ><%= link_to_unless(ParentObject.find_by_oid(parent_oid).nil?, parent_oid, parent_object_path(parent_oid)) %></td>
```
orangewolf: this is an extra query we probably need to refactor.

orangewolf (changes requested):

orangewolf:
@maxkadel consolidated this.  does this look ok?

orangewolf (approved):

PR #359, opened 2020-10-22
Scriptfix

mlooney:


README.md
```diff
@@ -126,7 +126,7 @@ aws_secret_access_key=YOUR_SECRET_ACCESS_KEY
 After the credentials have been set, you will need to export the following settings via the command line:
 
 ```bash
-export AWS_PROFILE=your-profile-name && export AWS_DEFAULT_REGION=us-east-1
+export AWS_PROFILE=your-profile-name AWS_DEFAULT_REGION=us-east-1
```
maxkadel, line 129: Are you sure this is right? I think you have to export them separately?

maxkadel (dismissed):
Looks good, just not sure about readme change

maxkadel (approved):

README.md
orangewolf, line 129: thanks! I didn't know you could do that.

spec/models/pyramidal_tiff_spec.rb
```diff
@@ -88,7 +88,7 @@
       .to_return(status: 200, body: File.open('spec/fixtures/images/sample_cmyk.tiff', 'rb'))
     ptiff_tmpdir = "spec/fixtures/images/ptiff_images/"
     allow(described_class).to receive(:new).and_return(ptf)
-    allow(ptf).to receive(:convert_to_ptiff).and_return(ptf.convert_to_ptiff("spec/fixtures/images/sample_cmyk.tiff", ptiff_tmpdir))
+    allow(ptf).to receive(:convert_to_ptiff).and_return(ptf.convert_to_ptiff(__FILE__, ptiff_tmpdir))
```
orangewolf, line 91: doesn't this mean you're trying to convert this file in to a ptiff now? how does that work?

vips/debian/README.source
```diff
@@ -0,0 +1,9 @@
+vips for Debian
```
orangewolf: can we move the vips files in to `ops` directory?

orangewolf (dismissed):
Looks good on the whole, one tiny change and one comment about the specs

spec/models/pyramidal_tiff_spec.rb
maxkadel: This is in the context of a test that asks "if I can't turn file x into a ptiff, does the application still behave appropriately" and the old "file x" can now be turned into a ptiff (good!) but now we need a new file that *can't* be turned into a ptiff in order for the test to pass.

fnibbit (approved):

PR #361, opened 2020-10-23
Datatables from passing

maxkadel:
- Pulled from the last passing version of code
- Re-merged in master
- Found failing tests (different ones! fun!)
- Poked the internet about testing AJAX stuff with RSpec 
- Figured out we weren't using Capybara for System spec if js isn't set to `true`, set it to `true`, re-ran specs, got legit failure!
  - Fixed legit failure


mlooney (approved):

spec/system/batch_process_spec.rb
```diff
@@ -62,15 +53,9 @@
...
-        end
-        within "td.count" do
-          expect(page).to have_content('5')
-        end
-        within "td.view" do
-          expect(page).to have_link('View', href: "/batch_processes/#{BatchProcess.last.id}")
-        end
+        expect(page).to have_link(BatchProcess.last.id.to_s, href: "/batch_processes/#{BatchProcess.last.id}")
+        expect(page).to have_content('5')
+        expect(page).to have_link('View', href: "/batch_processes/#{BatchProcess.last.id}")
```
alishaevn, line 58: I tried something similar to this in another test above but it didn't work for me. maybe the difference was adding `js:true`?

PR #365, opened 2020-10-26
Parent objects generated through the UI connected to batch process

maxkadel:


app/models/batch_process.rb
```diff
@@ -42,7 +42,8 @@ def file=(value)
   end
 
   def oid
-    self[:oid] = mets_doc.oid.to_i
+    return self[:oid] = mets_doc.oid.to_i unless self[:oid]
```
mlooney, line 45: avoid unless, refactor for clarity?

app/models/batch_process.rb
```diff
@@ -55,7 +56,7 @@ def mets_doc
 
   def oids
     return @oids ||= parsed_csv.entries.map { |r| r['oid'] } unless csv.nil?
```
mlooney, line 58: same as above

mlooney (changes requested):
see comments

app/controllers/parent_objects_controller.rb
```diff
@@ -102,6 +100,13 @@ def set_parent_object
       @parent_object = ParentObject.find(params[:id])
     end
 
+    def batch_process_of_one
+      @batch_process = BatchProcess.new(user: current_user, oid: @parent_object.oid)
```
mlooney, line 104: are instance variables needed here? will locals do?

mlooney (approved):

PR #368, opened 2020-10-27
Safe search to address bug

maxkadel:
Haven't been able to write a test to duplicate it, think it might be old parent objects?

When I deployed master of Management to collections-test earlier, the Batch Processes data table did not load, and got this error - https://app.honeybadger.io/projects/72674/faults/69213887 

I think this should hopefully address the bug, but I haven't been able to write a test that replicates the error yet. My best guess is that it is from old parent objects that were connected to Batch Processes, but not via the batch connection? I'm not entirely sure.

### DIFF
--- a/app/models/batch_process.rb
+++ b/app/models/batch_process.rb
@@ -42,7 +42,7 @@ class BatchProcess < ApplicationRecord
   end
 
   def oid
-    return self[:oid] = mets_doc.oid.to_i unless self[:oid]
+    return self[:oid] = mets_doc&.oid&.to_i unless self[:oid]
     self[:oid]
   end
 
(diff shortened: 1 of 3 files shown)


app/models/batch_process.rb
```diff
@@ -42,7 +42,7 @@ def file=(value)
   end
 
   def oid
-    return self[:oid] = mets_doc.oid.to_i unless self[:oid]
+    return self[:oid] = mets_doc&.oid&.to_i unless self[:oid]
```
alishaevn, line 45: 👍🏾

alishaevn (approved):
@maxkadel pulled this branch and the batch process data table loaded, but it loaded locally beforehand.

when jp & I were doing the data table work, we could get an error like this while testing initially. there's a method `oid` on the batch process model, but there's also a column of `oid` on each batch process. if I remember correctly, while debugging the test we would create a new batch process using the factory, but when calling `batch_process.oid`, we got the error that's showing in honeybadger.

when we changed the name of the method on the model, the conflict stopped happening. we did undo that change though, just wanted to verify that's what was causing the issue.

unsure if that's related to the parent objects or not.

approving this because it works locally for me and all the specs pass locally. but not merging in case the above information changes what you think is the cause of this issue. 🙂

PR #370, opened 2020-10-28
675 parent job detail page

maxkadel:
- Create a concern / module for Status, pulling from notifications
  - Include in parent object
- Add controller and routes for showing a given parent object's actions connected to a given batch process

image


app/models/concerns/statable.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+module Statable
+  extend ActiveSupport::Concern
+  def notes_for_batch_process(batch_process_id)
+    return if self.class == Integer
+    note_records = Notification.where(["params->>'batch_process_id' = :id and
+    params->>'parent_object_id' = :oid", { id: batch_process_id.to_s, oid:
+    oid.to_s }])
+    notes = {}
```
orangewolf: This could replace lines 10-12

```suggestion
    note_records.inject({}) { |i, n| i[n.params[:status]] = n.created_at; i }
```

app/models/concerns/statable.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  def status_for_batch_process(batch_process_id)
+    notes = notes_for_batch_process(batch_process_id)
+    if self.class == Integer
+      "Deleted or not created"
+    elsif self.class == ParentObject && notes["solr-indexed"]
+      "Complete"
+    else
+      "In progress or failed, who's to say?"
```
orangewolf: ?

app/views/batch_processes/show_parent.html.erb
```diff
@@ -0,0 +1,76 @@
+<h1>Parent Batch Process Detail</h1>
+<p id="notice"><%= notice %></p>
+<p>
+  <strong>Batch Process ID</strong>
+  <%= link_to @batch_process.id, batch_process_path(@batch_process) %>
+</p>
+<p>
+  <strong>Parent Object</strong>
+  <%= if @parent_object.class == ParentObject
```
orangewolf: how is parent object ever not a parent object? (this is not meant as a deep a philosophical question or riddle ;-) )

app/views/batch_processes/show_parent.html.erb
```diff
@@ -0,0 +1,76 @@
...
+  <thead class="thead-dark">
+    <tr>
+      <th scope="col"> Child object oid </th>
+      <th scope="col"> Last ptiff conversion </th>
+    </tr>
+  </thead>
+  <% if @parent_object&.class == ParentObject %>
+    <% @parent_object&.child_objects.each do |child| %>
+      <tr>
+        <td class="child_oid"><%= link_to_unless(child&.nil?, child&.oid, child_object_path(child)) || "(pending or deleted)" %></td>
```
orangewolf: `unless child&.nil?` is just a long way of saying `if child` isn't it?

orangewolf:
```suggestion
        <td class="child_oid"><%= child ? link_to(child.oid, child) : "(pending or deleted)" %></td>
```

orangewolf (changes requested):
This feels really solid Max. I don't get what the class == Integer stuff is about, maybe we need it, but if so I'd like to see at least a code comment about it and there are a few efficiency things that are probably worth looking at (inject is faster than alloc => map => dealloc for example)

app/models/concerns/statable.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+module Statable
+  extend ActiveSupport::Concern
+  def notes_for_batch_process(batch_process_id)
+    return if self.class == Integer
```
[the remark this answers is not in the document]
maxkadel: I wanted to have access to the parent object's oid, even when the parent object had been deleted, but in a way that was hopefully safe from injection? See line 63 in in the batch_processed_controller. So, I was calling the oid passed in the "parent object" if the parent object didn't exist.

On reflection, this is probably a bad way to do this. Maybe I can call it something like "requested_oid" to make it clear that it's a) a parameter passed in from the user b) should be an integer, and if it's not you should kill it with fire c) it's not a parent object

app/views/batch_processes/show_parent.html.erb
```diff
@@ -0,0 +1,91 @@
...
+<p>
+  <strong>Status</strong>
+  <%= "#{@parent_object.status_for_batch_process(@batch_process.id)}" %>
+</p>
+<p>
+  <strong>Duration</strong>
+  <%=  "#{@parent_object.duration_for_batch_process(@batch_process.id)}" %>
+</p>
+
+<table class="table table-bordered table-striped">
```
alishaevn, line 25: shouldn't this be a data table like the others?

app/views/batch_processes/show_parent.html.erb
```diff
@@ -0,0 +1,91 @@
...
+  </tr>
+</table>
+<table class="table table-bordered table-striped">
+  <thead class="thead-dark">
+    <tr>
+      <th scope="col"> Child object oid </th>
+      <th scope="col"> Last ptiff conversion </th>
+    </tr>
+  </thead>
+  <% if @parent_object&.class == ParentObject %>
```
alishaevn: shouldn't this be a data table like the others?

orangewolf: I agree with you @alishaevn, though for the sake of getting this in, I think we break that out to a separate PR

app/jobs/setup_metadata_job.rb
```diff
@@ -6,8 +6,8 @@ class SetupMetadataJob < ApplicationJob
   def perform(parent_object, current_batch_process)
     parent_object.current_batch_process = current_batch_process
     parent_object.generate_manifest = true
-    parent_object.default_fetch
-    parent_object.create_child_records
+    parent_object.default_fetch(current_batch_process)
+    parent_object.create_child_records(current_batch_process)
```
orangewolf, line 10: great addition

orangewolf (approved):

PR #375, opened 2020-10-29
Datatables error

maxkadel:


app/datatables/batch_process_datatable.rb
```diff
@@ -14,10 +14,10 @@ def view_columns
...
     # or in aliased_join_table.column_name format
     @view_columns ||= {
-      process_id: { source: "BatchProcess.id" },
-      user: { source: "BatchProcess.user_id" },
-      time: { source: "BatchProcess.created_at" },
-      size: { source: "BatchProcess.oid" },
+      process_id: { source: "BatchProcess.id", cond: :like, searchable: true },
+      user: { source: "BatchProcess.user_id", cond: :like, searchable: true },
+      time: { source: "BatchProcess.created_at", cond: :like, searchable: true },
+      size: { source: "BatchProcess.oid", cond: :like, searchable: true },
```
alishaevn, line 20: do you think this is part of the cause for the error?
I don't mind that they're here; we just left them off because these are the default values.

app/models/batch_process.rb
```diff
@@ -54,6 +50,10 @@ def mets_doc
     @mets_doc ||= MetsDocument.new(mets_xml) if mets_xml.present?
   end
 
+  def mets_oid
```
alishaevn, line 53: 👍🏾

alishaevn (approved):

app/datatables/batch_process_datatable.rb
maxkadel: hrm, possibly not, but I was getting an error locally that it couldn't find the search index or something, and this fixed it? And I thought we were seeing search wasn't working in production? Can't hurt might help?

PR #377, opened 2020-10-29
Batch detail datatables fresh

maxkadel:
- Add datatables for Batch Processes 
- Show parent object's status on batch process show page 

image

image

### DIFF
--- a/app/models/batch_connection.rb
+++ b/app/models/batch_connection.rb
@@ -3,4 +3,11 @@
 class BatchConnection < ApplicationRecord
   belongs_to :batch_process
   belongs_to :connectable, polymorphic: true
+
+  def child_object_count
+    return unless connectable_type == "ParentObject"
+    ParentObject.find(connectable_id)&.child_object_count
+  rescue ActiveRecord::RecordNotFound
+    nil
+  end
 end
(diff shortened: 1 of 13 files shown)


mlooney (approved):

app/models/batch_connection.rb
```diff
@@ -3,4 +3,17 @@
 class BatchConnection < ApplicationRecord
   belongs_to :batch_process
   belongs_to :connectable, polymorphic: true
+
+  def child_object_count
```
orangewolf, line 7: I think this method should exist only on the parent object it never needs to exist in this class. 

```ruby
batch_connection.connectable&.child_object_count
```

orangewolf (changes requested):
I haven't finished reviewing but I dont want this to get merged yet

app/models/batch_connection.rb
```diff
@@ -3,4 +3,17 @@
...
+    return unless connectable_type == "ParentObject"
+    ParentObject.find(connectable_id)&.child_object_count
+  rescue ActiveRecord::RecordNotFound
+    nil
+  end
+
+  def status
+    ParentObject.find(connectable_id).status_for_batch_process(batch_process_id)
```
orangewolf, line 15: you shouldn't ever need to do PO.find here.

`connection&.status_for_batch_process(batch_process_id) || "deleted"` makes more sense to me, would be a lot more efficient and might work in all cases

PR #379, opened 2020-10-30
Utilize connection more gracefully

maxkadel:
- They are more magic than I thought
- Removing spec for unneeded methods, please ignore code coverage if it complains about this, it is a liar.


orangewolf (approved):
Great work here =-)

PR #380, opened 2020-10-31
Batch page for child records

orangewolf:
Still needs 
- link to parent page
- link to access master
- spec fixes

### DIFF
--- a/app/controllers/batch_processes_controller.rb
+++ b/app/controllers/batch_processes_controller.rb
@@ -1,8 +1,11 @@
 # frozen_string_literal: true
 
 class BatchProcessesController < ApplicationController
-  before_action :set_batch_process, only: [:show, :edit, :update, :destroy, :download, :download_csv, :download_xml, :show_parent]
-  before_action :set_parent_object, :find_notes, :find_failures, only: [:show_parent]
+  before_action :set_batch_process, only: [:show, :edit, :update, :destroy, :download, :download_csv, :download_xml, :show_parent, :show_child]
+  before_action :set_parent_object, only: [:show_parent, :show_child]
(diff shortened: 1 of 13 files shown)


app/controllers/batch_processes_controller.rb
```diff
@@ -1,8 +1,11 @@
 # frozen_string_literal: true
 
 class BatchProcessesController < ApplicationController
-  before_action :set_batch_process, only: [:show, :edit, :update, :destroy, :download, :download_csv, :download_xml, :show_parent]
-  before_action :set_parent_object, :find_notes, :find_failures, only: [:show_parent]
+  before_action :set_batch_process, only: [:show, :edit, :update, :destroy, :download, :download_csv, :download_xml, :show_parent, :show_child]
```
maxkadel, line 4: I like this reorganization

app/controllers/batch_processes_controller.rb
```diff
@@ -64,8 +69,12 @@ def set_batch_process
 
     def set_parent_object
       @parent_object = ParentObject.find(params[:oid])
```
maxkadel: I think we still need this if we want to show historical batch process data for parent objects that have since been deleted

app/models/parent_object.rb
```diff
@@ -31,6 +31,14 @@ def initialize(attributes = nil)
     self.use_ladybird = true
   end
 
+  def start_states
```
maxkadel, line 34: I like this approach a lot

maxkadel (changes requested):
Looks good, like the refactor of the starting and ending states particularly.

maxkadel (approved):

PR #381, opened 2020-11-02
Adds batch status to batch pages

maxkadel:
- Status not currently represented in the DB so not sortable
- Tried just adding a column, but needs more work wrt when to update status before that's ready

### DIFF
--- a/app/models/batch_process.rb
+++ b/app/models/batch_process.rb
@@ -89,4 +89,38 @@ class BatchProcess < ApplicationRecord
       refresh_metadata_cloud_mets
     end
   end
+
+  def batch_status
+    if single_status
+      single_status
+    elsif statuses[:failed] != 0
+      percent = ((statuses[:failed] / statuses[:total]) * 100).round
+      "#{percent}% of parent objects have a failure."
+    else
+      "Batch status unknown"
+    end
+  end
+
+  def single_status
+    if statuses[:in_progress] / statuses[:total] == 1
+      "Batch in progress - no failures"
+    elsif statuses[:complete] / statuses[:total] == 1
+      "Batch complete"
+    elsif statuses[:failed] / statuses[:total] == 1
+      "Batch failed"
+    end
+  end
+
+  def statuses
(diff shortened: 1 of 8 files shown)


app/models/batch_process.rb
```diff
@@ -89,4 +89,38 @@ def refresh_metadata_cloud
...
+    elsif statuses[:failed] != 0
+      percent = ((statuses[:failed] / statuses[:total]) * 100).round
+      "#{percent}% of parent objects have a failure."
+    else
+      "Batch status unknown"
+    end
+  end
+
+  def single_status
+    if statuses[:in_progress] / statuses[:total] == 1
```
mark-dce, line 105: I think you might want to capture the statuses one time here - I think the way you have this structured, you might be calling batch_connections.map up to 6 times.

```
current_status = statuses
if current_status[:in_progress] / current_status[:total] == 1
... etc.
```

app/models/batch_process.rb
```diff
@@ -89,4 +89,38 @@ def refresh_metadata_cloud
...
+      "Batch in progress - no failures"
+    elsif statuses[:complete] / statuses[:total] == 1
+      "Batch complete"
+    elsif statuses[:failed] / statuses[:total] == 1
+      "Batch failed"
+    end
```
mark-dce, line 111: Let's add an else here to catch any unexpected states

```
else
    "Unexpected state"

**Review by mlooney — APPROVED** [review]

**Review by mark-dce — CHANGES_REQUESTED** [review]
This looks great.  I've marked two small changes in calculating status that I think might make the status calculation faster and catch a possible future error state that we "shouldn't " be able to reach, but might at some point in the future if someone refactors statuses.

File: app/models/batch_process.rb [hunk 3]
```diff
@@ -89,4 +89,38 @@ def refresh_metadata_cloud
...
+  end
+
+  def statuses
+    connected_statuses = batch_connections.map do |batch_connection|
```
orangewolf, line 115: this method can be refactored for speed up, I think significantly.

PR #387, opened 2020-11-04
Pdf generation concern

martinlovell:
Start of PDF generation.

Getting it in so others can use ParentObject.remote_pdf_path

### DIFF
--- a/app/models/concerns/pdf_representable.rb
+++ b/app/models/concerns/pdf_representable.rb
@@ -66,7 +66,7 @@ module PdfRepresentable
       container_information = extract_container_information(authoritative_json)
       properties["Container information"] = container_information if container_information
       properties["Generated"] = generated
-      properties["Terms of Use"] = "https://guides.library.yale.edu/about/policies/use-of-materials"
+      properties["Terms of Use"] = "https://guides.library.yale.edu/about/policies/access"
       properties["View in DL"] = dl_show_url.to_s
       reshape_properties properties
     end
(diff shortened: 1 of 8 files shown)


app/models/concerns/pdf_representable.rb
```diff
@@ -0,0 +1,106 @@
+# frozen_string_literal: true
+
+module PdfRepresentable
+  extend ActiveSupport::Concern
+
+  def generate_pdf
+    raise "No authoritative_json to create PDF for #{oid}" unless authoritative_json
+    # temp_json_file = Tempfile.new("#{oid}_pdf_json")
```
mlooney: can we clean this up? or is there value here?

app/models/concerns/pdf_representable.rb
```diff
@@ -0,0 +1,106 @@
...
+  def remote_pdf_path
+    "pdfs/#{Partridge::Pairtree.oid_to_pairtree(oid)}/#{oid}.pdf"
+  end
+
+  def pdf_generator_json
+    generated = Time.now.utc.to_s
+    title = extract_flat_field_value(authoritative_json, "title", "No Title")
+    properties = pdf_properties title, generated
+    children = child_pages
+    return nil if children.empty?
```
mlooney: consider returning empty hash instead of nil, caller can do the x.is_empty? if they feel like it

app/models/concerns/pdf_representable.rb
martinlovell: Max and I cleaned this up.

app/models/concerns/pdf_representable.rb
martinlovell: Changed it to return json with emtpy child hash even when there are no children.

maxkadel (approved):

app/models/concerns/pdf_representable.rb
martinlovell: Updated jar to generate PDF with cover page if there are no children.

PR #389, opened 2020-11-05
Turn parent detail page into a data table

alishaevn:
Co-authored by: JP Engstrom <jp@notch8.com>

Turned parent detail page into a datatable. Added specs to batch process child detail
# Ticket
https://app.zenhub.com/workspaces/yul-dc-5e50083561915b11fc9e5850/issues/yalelibrary/yul-dc/675

### DIFF
--- a/app/models/child_object.rb
+++ b/app/models/child_object.rb
@@ -43,6 +43,7 @@ class ChildObject < ApplicationRecord
   def access_master_path
     return @access_master_path if @access_master_path
     image_mount = ENV['ACCESS_MASTER_MOUNT'] || "data"
+    # TODO(alishaevn): the "pairtree_path" variable is coming back as "nil" in the batch process child detail spec
     pairtree_path = Partridge::Pairtree.oid_to_pairtree(oid)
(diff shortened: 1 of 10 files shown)


app/models/child_object.rb
```diff
@@ -43,6 +43,7 @@ def access_master_url
   def access_master_path
     return @access_master_path if @access_master_path
     image_mount = ENV['ACCESS_MASTER_MOUNT'] || "data"
+    # TODO(alishaevn): the "pairtree_path" variable is coming back as "nil" in the batch process child detail spec
```
maxkadel, line 46: JP and I paired on diagnosing this, fixing the test, and writing a bug ticket, so I don't think we still need this TODO here (see https://github.com/yalelibrary/YUL-DC/issues/827)

orangewolf (approved):

app/models/child_object.rb
jpengst: Ah yes, good catch. Sorry, forgot about that comment.

PR #391, opened 2020-11-09
Queue PDF job whenever manifest is changed.

martinlovell:
Queues PDF Job whenever manifest is generated.  (Relies on the fact that the manifest is only generated when necessary.  Manifest should be regenerated in the same situations when the PDF should be regenerated.)


maxkadel (approved):

maxkadel (changes requested):
Sorry, making a slight mess of this PR review - can we add the batch_connection & batch_process so we can use the same notification process to report back on creating PDFs? They can be non-required arguments

maxkadel (approved):
Thanks!

PR #392, opened 2020-11-09
Capture pdf gen output

martinlovell:
Before: java app output printed to stdout
After: java app output captured.  Added to error message when non-zero response code is returned


mlooney (changes requested):
refer to app/models/pyramidal_tiff.rb:74 for a way to capture both stderr & stdout

maxkadel (approved):

mlooney (approved):

PR #393, opened 2020-11-09
Add rendering to iiif presentation

maxkadel:
Adds link to not-yet-connected pdfs service in manifest, allowing it to display in Universal Viewer. 

Connected to https://github.com/yalelibrary/YUL-DC/issues/764

image


mlooney (approved):
looks good, I assume the templates in camerata will be updated to include this new env var :)

PR #397, opened 2020-11-10
Restrict login based on list of uids on S3

martinlovell:
See README for details: https://github.com/yalelibrary/yul-dc-management/tree/803_restrict_access#authorization

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -9,3 +9,4 @@ tmp
 yarn.lock
 .github
 data_samples
+config/cas_users.csv
--- a/.gitignore
+++ b/.gitignore
@@ -24,13 +24,6 @@
 /data_samples/*
 /data/*
 !/data/.keep
-
-/public/assets
-.byebug_history
-
-# Ignore master key for decrypting credentials and more.
-/config/master.key
-
 /public/packs
 /public/packs-test
 /node_modules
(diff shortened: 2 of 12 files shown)


martinlovell:
The CSV file has been uploaded to the buckets for each environment.

maxkadel (changes requested):
I think this is an excellent start, and might be worth merging with minor tweaks as a start, but I'd want Rob's feedback first. 

- If we use the rake task on an ongoing basis, I think we should gitignore (and maybe Dockerignore?) the `config/cas_users.csv` file so that it isn't accidentally committed to version control.
- I believe the agreement was to initially seed the environments with the csv file, but then after that to use the Rails native CRUD within the application to manage users, meaning that part of the ticket is to create a user controller, model, etc. 
- Right now, I think if we added the CRUD and used it to add a new member to the application, on the next deployment that user would be removed. I like the idea of periodically being able to normalize / purge the user list (and think it was a requirement?), but I don't think we want to do so on each deploy. 
- There are no tests for this change, and *especially* since this is a security change, I'd love to see tests that verify at least the changed behavior in the omniauth_callbacks_controller

orangewolf (approved):
This skips the admin ui completely and assumes that you'll update the csv any time you want to add ore remove someone from the list.  Honestly that feels cumbersome over time as a dev task and not something the users can do for themselves, but this solves the immediate security need. Let's move forward here, then build the system @mikeapp actually wants for the next swing.

maxkadel (approved):

PR #416, opened 2020-11-24
Ingest event again

maxkadel:


app/controllers/batch_processes_controller.rb
```diff
@@ -78,16 +78,16 @@ def set_parent_object
 
     def set_child_object
       @child_object = ChildObject.find(params[:child_oid])
-      @notes = @child_object.notes_for_batch_process(@batch_process.id)
-      @failures = @child_object.latest_failure(@batch_process.id)
+      @notes = @child_object.notes_for_batch_process(@batch_process)
```
orangewolf, line 81: this is a very welcome change

app/models/child_object.rb
```diff
@@ -22,8 +22,14 @@ def check_for_size_and_file
     width_and_height(remote_metadata)
   end
 
-  def processing_event(message, status = 'info', current_batch_process = parent_object&.current_batch_process, _current_batch_connection = parent_object&.current_batch_connection)
-    IngestNotification.with(parent_object_id: parent_object&.id, child_object_id: id, status: status, reason: message, batch_process_id: current_batch_process&.id).deliver(User.first)
+  def processing_event(message, status = 'info', _current_batch_process = parent_object&.current_batch_process, current_batch_connection = parent_object&.current_batch_connection)
+    return unless current_batch_connection
+    ie = IngestEvent.new(
```
orangewolf: just do create! instead of new and then save!, that way you do not alloc and then free a variable from memory

orangewolf (changes requested):
one super small change and the bigger question - what are we hoping to gain from this code change. we're adding a new model to the code and (I assume) removing one eventually?

orangewolf (approved):
thanks for the clarification

maxkadel:
> one super small change and the bigger question - what are we hoping to gain from this code change. we're adding a new model to the code and (I assume) removing one eventually?

For posterity: 
My thinking was that
1. I have an untested hypothesis (part of why I'm about to play with it on collections-test) that searching the fields directly, rather than within the json parameters of the notifications, will be faster
2. We're creating lots of notifications that are not necessarily actionable by a single user
3. This frees us up to only notify the user who created the batch process, and only tell them what they are more likely to care about (the final result, when they've all reach some kind of "failed or other" end state), and for those not to get mixed in with the status notifications. I was going to start on that but I felt like I needed some more input on what people care about and how to determine it

PR #424, opened 2020-12-02
Always call metadata cloud with a version

maxkadel:
When making a call to the metadatacloud api we should always include the version.


FCRodriguez7:
We made changes to have the MC use mapping from version 1.0.1.  New and future mapping will be placed on the API version of 1.0.2.

FCRodriguez7:
Additionally, work: https://github.com/yalelibrary/yul-dc-management/pull/426

fnibbit (approved):

PR #426, opened 2020-12-02
683 update solr to MC101

FCRodriguez7:
We have changed fields in the latest MC, need to change SOLR code to match. May be blocked by versioning work?
- [x] author -> creator
- [x] IdentifierShelfMark -> callNumber (LB and ILS only? Don't think that this exists in Aspace)
- [ ] PublicationDate -> date or createDate (group is working out date fields)
- [x] geoSubject -> subjectGeographic
- [x] partOf -> sourceTitle (need to test this, correct for LB, not sure how it affects aspace/voyager)
- [x] change box to [rename] containerGrouping (for consistency with aspace and voyager)
- [x] check for other changes to MC
     - [x] publicationPlace_ssim: json_to_index["creationPlace"]
     - [x] publicationPlace_tesim: json_to_index["creationPlace"]
     - [x] references_ssim: json_to_index["preferredCitation"]
     - [x] references_tesim: json_to_index["preferredCitation"]

Update SOLR mappings in blacklight to match.
Blacklight labels do not have to correspond based on these updates, that will be different ticket work as some will be different for reading comprehension purposes


app/models/parent_object.rb
```diff
@@ -127,7 +127,7 @@ def ladybird_json=(lb_record)
     return lb_record if lb_record.blank?
     self.last_ladybird_update = DateTime.current
     return unless use_ladybird
-    self.bib = lb_record["orbisBibId"] || lb_record["orbisRecord"]
+    self.bib = lb_record["orbisBibId"]
```
maxkadel, line 130: If we have to change all our fixture objects to accommodate this change, does that maybe indicate we should leave the "or" statements for now, until we've re-fetched everything from the new version of the metadata cloud in the production-like environments?

Then we can re-fetch all our fixtures from the metadata cloud, make sure everything is still passing, and do the updates to the fixtures in a separate PR?

maxkadel (changes requested):

maxkadel (approved):
Approved, with a request that you look at my PR to fully update all the fixture files.

PR #430, opened 2020-12-03
Use database upsert_all method to batch create child objects

maxkadel:
* When trying to create many objects locally, noticed that ChildObject creation seemed to be a blocker for fast import
* *Before* Used benchmark method on test of child object creation on ParentObject with 156 children and got total times of ~8 seconds. When tried on ParentObject with ~1300 children it would time out. 
* *After* With larger ParentObject it would not time out, and took < 1 second total. 
* I believe it is ok that for the batch creation of child objects (when first creating a parent object), we are not running callbacks on the individual child object, since the child object background jobs are called from within the parent object background jobs. Put in a reminder above the "upsert_all" call in case this changes in future.
* Added default values to creation_date and updated_date, which mirrors current functionality but moves it to the database, so that we would not have to put these timestamps in the upsert_all method, which would run the risk of overriding older creation dates for re-ingested ParentObject children.

### DIFF
--- a/app/models/parent_object.rb
+++ b/app/models/parent_object.rb
@@ -42,19 +42,27 @@ class ParentObject < ApplicationRecord # rubocop:disable Metrics/ClassLength
     ['solr-indexed', 'pdf-generated']
   end
 
-  def create_child_records(_current_batch_process, _current_batch_connection = current_batch_connection)
+  def create_child_records
+    return unless ladybird_json
+    return self.child_object_count = 0 if ladybird_json["children"].empty?
+    # Note - the upsert_all method skips ActiveRecord callbacks, and is entirely
+    # database driven. This also makes object creation much faster.
+    ChildObject.upsert_all(array_of_child_hashes)
+    self.child_object_count = child_objects.size
+  end
+
+  def array_of_child_hashes
     return unless ladybird_json
     ladybird_json["children"].map.with_index(1) do |child_record, index|
-      next if child_object_ids.include?(child_record["oid"])
-      child_objects.build(
+      {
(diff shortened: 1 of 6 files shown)


app/models/parent_object.rb
```diff
@@ -42,19 +42,27 @@ def finished_states
...
         # Ladybird has only one field for both order label (7v, etc.), and descriptive captions ("Mozart at the Keyboard")
         # For the first iteration we will map this field to label
         label: child_record["caption"],
-        order: index
-      )
+        order: index,
+        parent_object_oid: oid
+      }
     end
-    self.child_object_count = child_objects.size
```
orangewolf, line 57: we should remove this counter cache from the db if we are not using it. could easily be a separate chore ticket though.

orangewolf (approved):

app/models/parent_object.rb
```diff
@@ -42,19 +42,27 @@ def finished_states
...
   end
 
-  def create_child_records(_current_batch_process, _current_batch_connection = current_batch_connection)
+  def create_child_records
+    return unless ladybird_json
+    return self.child_object_count = 0 if ladybird_json["children"].empty?
+    # Note - the upsert_all method skips ActiveRecord callbacks, and is entirely
+    # database driven. This also makes object creation much faster.
+    ChildObject.upsert_all(array_of_child_hashes)
+    self.child_object_count = child_objects.size
```
maxkadel, line 51: The counter is moved here - just a git diff quirk. We're indexing it to Solr in order to display how many images are available for a parent object in Blacklight.

PR #438, opened 2020-12-10
Add iiif to children

maxkadel:
- Add to form for editing child objects
- Allow through controller
- Set options for form
- Add to iiif presentation canvas if saved and not an empty string


mikeapp (approved):

mikeapp:
These changes look good in and of themselves, I don't know what's causing the test failure.

PR #441, opened 2020-12-11
Set parent thumbnail

maxkadel:
* Currently, the representative thumbnail defaults to the first image. Often this is a good representation of the object, but sometimes it is not (e.g., the front of a library binding). I would like to be able to select which child image is appropriate to represent the parent object in the catalog via the Management UI 

This PR
- Shows the representative thumbnail on the parent object's page in the management UI
image
- Adds a link to the page where you can select the representative child / representative thumbnail from the ParentObject edit page
image
- Creates the page where the appropriate thumbnail can be selected 
image
image

Further work includes:
- ~~paginating or only allowing the user to select from the first few pages, in order to not overwhelm the cantaloupe server / make the page unusably slow~~
- ~~UI enhancements (should the "select different thumbnail" link only be on the parent object show page? on both the show page and the edit page?)~~
- ~~Further tests for selection page~~


mikeapp:
It would be useful to store the OID or child reference.  There's the notion of both a Manifest thumbnail and a "start Canvas" in IIIF, but we'd need to generate JSON for an IIIF image service and the Canvas URI/`@id` , both of which will be possible if we have the OID.   Also, it might be good to check that the child for the thumbnail hasn't been deleted/reparented.  

Edit: missed that you store both, yay!  

Ticket for possible IIIF work: https://app.zenhub.com/workspaces/yul-dc-5e50083561915b11fc9e5850/issues/yalelibrary/yul-dc/923

config/routes.rb
```diff
@@ -18,6 +18,7 @@
     end
     member do
       post :update_metadata
+      get '/select_thumbnail', to: 'parent_objects#select_thumbnail', as: :select_thumbnail
```
orangewolf: use of restful helpers instead of hard coding routes is preferred https://guides.rubyonrails.org/routing.html#adding-member-routes

```suggestion
      get :select_thumbnail
```

spec/fixtures/medium_short_fixture_ids.csv
```diff
@@ -0,0 +1,28 @@
+oid
```
orangewolf: medium short? feels like there has to be a more purpose focused name for this file

orangewolf (changes requested):
I would say this should not be merged unless pagination is put in. doing so is pretty easy and it will keep large parent objects from tying up production processes because someone clicked the button

orangewolf:
Looking good, though I admit I'm a little unclear where the requirements come from or if they are well described

spec/fixtures/medium_short_fixture_ids.csv
maxkadel: Agree, and should be in a different PR as well

app/models/parent_object.rb
```diff
@@ -215,9 +215,20 @@ def ready_for_manifest?
     !child_objects.pluck(:width).include?(nil)
   end
 
-  def representative_thumbnail
-    oid = child_objects.where(order: 1)&.first&.oid
-    "#{ENV['IIIF_IMAGE_BASE_URL']}/2/#{oid}/full/!200,200/0/default.jpg"
+  def representative_child
+    if representative_child_oid
+      begin
```
orangewolf: use find_by_id instead here and skip the rescue. that way you only need the else block

```
@representative_child = if representative_child_oid
  child_objects.find_by(id: representative_child_oid)
end
@representative_child ||= child_objects.where(order: 1)&.first

orangewolf (changes requested):
small optimization - this is looking really great

app/models/parent_object.rb
maxkadel: Cool, Mark and I were thinking there must be a way but hadn't hit on it, thanks for that.

orangewolf (approved):

PR #446, opened 2020-12-17
Does not use rails helpers to set worker_count and worker_queue

maxkadel:
Co-Authored-By: Rob Kaufman <rob@notch8.com>


orangewolf (approved):

maxkadel:
Labeling a feature to make clear that it goes with the new Camerata v4.0.0

PR #453, opened 2020-12-22
Update goobi expectations

maxkadel:
Connected to https://github.com/yalelibrary/YUL-DC/issues/957


maxkadel:
replaced by rebased one

PR #455, opened 2020-12-23
Refactor ParentObjects - Will not always have Ladybird record in future

maxkadel:
As we move toward ingesting new objects directly through Goobi, we need to change the software expectations, and not assume that there will be Ladybird metadata available.

This PR makes it so that we do not always first pull the Ladybird metadata prior to pulling Voyager or ArchiveSpace metadata. 

If the information needed to create a Voyager or ArchiveSpace url is not available, and the application tries to generate the URL in order to pull the metadata, it will raise an error. 

### Future work 
  - Add validation via the UI to make it clear that if the source of authority is not Ladybird, and the ParentObject isn't coming through Goobi, you have to include additional identifiers in order to fetch the metadata (ArchiveSpace URI or Voyager bib id and barcode if applicable)
  - Decide how to use the identifier given by Goobi to retrieve both ArchiveSpace and Voyager records and update code accordingly

### DIFF
--- a/app/models/batch_process.rb
+++ b/app/models/batch_process.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class BatchProcess < ApplicationRecord
+class BatchProcess < ApplicationRecord # rubocop:disable Metrics/ClassLength
   attr_reader :file
   after_create :refresh_metadata_cloud
   before_create :mets_oid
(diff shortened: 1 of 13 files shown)


mlooney (approved):

maxkadel:
Does not appropriately assign visibility from Ladybird currently

PR #460, opened 2021-01-11
Update code for editing rights statements on parent objects

maxkadel:
- Don't have the background job error out if no ladybird_json
- Pass the rights statement through the controller. D'oh. Going on break problems.

### DIFF
--- a/app/jobs/add_rights_to_parent_objects_job.rb
+++ b/app/jobs/add_rights_to_parent_objects_job.rb
@@ -11,7 +11,7 @@ class AddRightsToParentObjectsJob < ApplicationJob
 
   def perform(*_args)
     ParentObject.where(rights_statement: nil).find_each do |parent|
-      parent.rights_statement = parent.ladybird_json["rights"]&.first if parent.ladybird_json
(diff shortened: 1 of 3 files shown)


app/jobs/add_rights_to_parent_objects_job.rb
```diff
@@ -11,7 +11,7 @@ def default_priority
 
   def perform(*_args)
     ParentObject.where(rights_statement: nil).find_each do |parent|
-      parent.rights_statement = parent.ladybird_json["rights"]&.first
+      parent.rights_statement = parent.ladybird_json["rights"]&.first if parent.ladybird_json
```
orangewolf:
```suggestion
      parent.rights_statement = parent.ladybird_json&.[]("rights")&.first
```

orangewolf: to me, this is more consistent (&. twice instead of a mix of &. and if)

orangewolf (approved):
minor code consistency thing. no need fo re-review

PR #462, opened 2021-01-11
Generate child objects and ptiffs from Goobi imports

maxkadel:
Co-authored-by: Rachel Kadel <rkadel@curationexperts.com>


app/models/parent_object.rb
```diff
@@ -13,6 +13,7 @@ class ParentObject < ApplicationRecord # rubocop:disable Metrics/ClassLength
   has_many :batch_processes, through: :batch_connections
   belongs_to :authoritative_metadata_source, class_name: "MetadataSource"
   attr_accessor :metadata_update
+  attr_accessor :from_mets
```
orangewolf: I  think it might be valuable to know that a record came from mets permanently (aka a db field) in stead of transiently. If that's the case, then swapping this for a property on the parent object and then changing it so its set on create instead of in the background jobs might make sense?

app/jobs/generate_ptiff_from_mets_job.rb
```diff
@@ -0,0 +1,19 @@
...
+class GeneratePtiffFromMetsJob < ApplicationJob
+  queue_as :ptiff
+
+  def default_priority
+    10
+  end
+
+  def perform(child_object, current_batch_process, current_batch_connection = child_object.parent_object.current_batch_connection)
+    child_object.parent_object.current_batch_process = current_batch_process
+    child_object.parent_object.current_batch_connection = current_batch_connection
```
orangewolf: there is a lot of duplication here. I want to respect your comment about drying it up after this gets in, but I wonder if a quick toweling off of the jobs now might be worth while. If the broader structural change of making "from_mets" a parent_object property happens (see below) then this could simply be collapsed in to the original job again.  We might want to separate concerns from there, or split out a callback, but that would seem more like refactoring to me.

app/jobs/setup_metadata_from_mets_job.rb
```diff
@@ -0,0 +1,22 @@
...
+
+class SetupMetadataFromMetsJob < ApplicationJob
+  queue_as :metadata
+
+  def perform(parent_object, current_batch_process, current_batch_connection = parent_object.current_batch_connection)
+    parent_object.current_batch_process = current_batch_process
+    parent_object.current_batch_connection = current_batch_connection
+    parent_object.generate_manifest = true
+    parent_object.default_fetch(current_batch_process, current_batch_connection)
+    parent_object.create_child_records_from_mets
```
orangewolf: if parent object gets permanent awareness of its "from_mets-ness" then I think `create_child_records` can just call `create_child_records_from_mets` or `create_child_records_from_ladybird` intelligently inside the parent object and this job no longer needs to happen separately =-)

orangewolf (changes requested):

orangewolf:
this is excellent code navigating a complex refactor. Good work.  I'm suggesting a change (making from_mets a property instead of an accessor) that I think simplifies things further. I may be missing a reason to not do that, in which case, of corse feel free to override =-)

maxkadel:
> this is excellent code navigating a complex refactor. Good work. I'm suggesting a change (making from_mets a property instead of an accessor) that I think simplifies things further. I may be missing a reason to not do that, in which case, of corse feel free to override =-)

Thanks - I'll try that out. I think the tricky thing to separate is things that should be done only the first time a parent or child object is created, such as copying the images from the Goobi package to the access master pair-tree.

orangewolf (approved):

PR #464, opened 2021-01-14
Make children column correctly find parent object after joining it to…

orangewolf:
… the datatable


maxkadel:
Locally, this fails `rspec ./spec/system/batch_process_detail_spec.rb:54`, which tests for the behavior if a ParentObject is imported and then subsequently deleted, which has been given as a requirement in the past. Adding this join means that if the ParentObject has been deleted, it will not be listed in the datatable.

orangewolf:
Thanks @maxkadel! I didn't think this little corner had or needed coverage, and I was 100% wrong about this. A bug would have been seen in production if it had not been caught by this test.

fnibbit (approved):

PR #468, opened 2021-01-19
Do not copy to Access Master pair-tree

maxkadel:
* In order to verify the rest of the Goobi ingest process, do not try to copy to the access master pair-tree, which is currently read only


maxkadel:
Have write access to directory now, still-relevant parts in new pull request

PR #480, opened 2021-02-02
Export all child oids connected to a given list of parent objects

maxkadel:
Add new functionality to Batch Processes
  * Select "export child oids" from dropdown (see first screenshot)
  * User uploads a CSV of Parent Object oids, with the same formatting as the Parent Object creation csv
  * In a background job, the output / created CSV is generated and saved to the database 
  * The user can then go to the batch process show page and download the created file (see second screenshot)
  * This PR also refactors the creation of parent objects from a CSV upload to a background job, to make CSV upload faster and more immediately clear whether it's been successful or not. This necessitated a number of test expectation changes.
image
image

### DIFF
--- a/app/jobs/generate_output_csv_job.rb
+++ b/app/jobs/generate_output_csv_job.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+class GenerateOutputCsvJob < ApplicationJob
+  queue_as :default
+
+  def perform(batch_process)
+    batch_process.output_csv
+  rescue => e
+    parent_object.processing_event("Setup job failed to save: #{e.message}", "failed", current_batch_process, current_batch_connection)
+    raise # this reraises the error after we document it
+  end
+end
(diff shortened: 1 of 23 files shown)


app/jobs/generate_output_csv_job.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class GenerateOutputCsvJob < ApplicationJob
+  queue_as :default
+
```
orangewolf, line 5: default priority?

app/models/batch_process.rb
```diff
@@ -47,6 +51,32 @@ def oids
     @oids ||= [oid]
   end
 
+  def output_csv
```
orangewolf: thoughts on making this (and created_file_name) a concern? this would mostly be the idea that CsvExportable is a set of features that can be kept together outside the primary model and not necessarily for reuse at this moment, so I'd consider it a tidying thing and not a structural one and definitely deep in the realm of opinion.

orangewolf (commented):
overall this looks great. couple of small things that I could be fine if someone else wants to approve and move forward.

orangewolf (approved):

PR #488, opened 2021-02-04
Excel utf 8 fix

maxkadel:


maxkadel:
nope - other version actually worker

PR #494, opened 2021-02-05
If no label is provided in metadata, use blank string in iiif manifest

martinlovell:


mikeapp (dismissed):

mikeapp:
Looks like the failed test has no Canvases

maxkadel (approved):

PR #498, opened 2021-02-05
Test reflecting encoding problem

maxkadel:
Added a BOM to the sent file to try to get Excel to behave, but that seems to have broken subsequent ingest of the same updated file. Test failing for the right reason.


mikeapp:
`U-FEFF` appears to be the BOM for UTF-16, but we're sending UTF-8?
See https://en.wikipedia.org/wiki/Byte_order_mark#UTF-8

maxkadel:
> `U-FEFF` appears to be the BOM for UTF-16, but we're sending UTF-8?
> See https://en.wikipedia.org/wiki/Byte_order_mark#UTF-8

Fair point, you're right, but when I try using the UTF-8 one instead of the UTF-16 it also fails to re-upload. The wrong BOM came from this thread - https://stackoverflow.com/questions/30368173/ruby-how-to-generate-csv-files-that-has-excel-friendly-encoding 

Aaaaaand the real answer has come from the far less popular thread here - https://stackoverflow.com/questions/57032550/what-does-rbbomutf-8-mean-in-csv-open-in-ruby

martinlovell (approved):

PR #509, opened 2021-02-11
Adds expects to ensure page is loaded and submitted when creating new ParentObject

martinlovell:
This fixes some timing issues with capybara and the new parent object form.
It waits (by using expect) until the form loads before filling out the Oid field, to ensure it's not still on the parent object list page which now has an Oid input.
Waits (with expect) after submitting form for response from server to ensure that the form submission was processed.

Used shell script to test multiple runs:
```
$ cat rununtilfail.sh
run=0
$@
while [ $? -eq 0 ]; do
    run=$((run+1))
    echo Run number $run
    $@
done
```

run with:
```
sh rununtilfail.sh rspec ./spec/system/batch_process_spec.rb:39
```

### DIFF
--- a/spec/system/batch_process_spec.rb
+++ b/spec/system/batch_process_spec.rb
@@ -27,9 +27,11 @@ RSpec.describe BatchProcess, type: :system, prep_metadata_sources: true, js: tru
       stub_metadata_cloud("16057779")
       visit parent_objects_path
       click_on("New Parent Object")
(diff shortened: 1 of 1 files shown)


maxkadel (approved):

maxkadel (changes requested):
Could you add a comment to that "expects" to note that it's forcing the correct page load, so it's not removed without realizing that?

maxkadel (approved):

PR #511, opened 2021-02-11
Url from xml check

maxkadel:


dl-maura (approved):
Looks good and worked in test

PR #514, opened 2021-02-11
Add directory scanner with url trigger

martinlovell:
Adds directory scanner which looks for XML files and queues them for the import process.

Scanning will use GOOBI_SCAN_DIRECTORIES which contains a comma separated list of directories to scan.  If GOOBI_SCAN_DIRECTORIES is not set, it will use /GOOBI_MOUNT/dcs as a fallback.

Done files are written when the Batch is successfully created and the first job in the import process is triggered.  Once the batch is created, errors further down in the process can be seen through management batch reporting.  This is, in part, to prevent the same failing import to run repeatedly and fill the system with jobs/batches.  The delayed job system will continue to retry the original job.

When the scanner starts the process of creating the batch and saving it, it creates a progress file to prevent other scanners from starting the process on the same mets file / directory.  If a progress file is older than 24 hours old, it is deleted and reprocessed.  (Progress files should typically be around for only a few minutes.  If we see older progress files, it means that the system had problems validating the mets or something else went wrong when setting up the batch and jobs.)

The scanner is not scheduled yet.  For testing purposes, there is a URL to trigger the scan.  Once we tested it, we can set it up to schedule the scan.  The scan is done with a simple job so that the mounts are available.

### DIFF
--- a/app/jobs/mets_directory_scan_job.rb
+++ b/app/jobs/mets_directory_scan_job.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+# Very simple job.  Scan needs to occur in a job so that the mounts are available
+class MetsDirectoryScanJob < ApplicationJob
+  queue_as :default
+
+  def default_priority
+    100
+  end
+
+  def perform
+    MetsDirectoryScanner.perform_scan
+  end
+end
--- a/app/lib/mets_directory_scanner.rb
+++ b/app/lib/mets_directory_scanner.rb
@@ -33,9 +33,7 @@ class MetsDirectoryScanner
       tempfile: File.new(xml_file_path.to_s)
     )
     begin
-      batch_process = BatchProcess.new(batch_action: 'create parent objects')
-      # set file with setter to trigger reading
-      batch_process.file = file
+      batch_process = BatchProcess.new(batch_action: 'create parent objects', user: system_user, file: file)
       if batch_process.save!
         # Jobs have been kicked off and batch job has been created, so we'll mark it as done and
         # check for errors in management.
@@ -61,5 +59,14 @@ class MetsDirectoryScanner
 
   def self.scan_directories
     ENV['GOOBI_SCAN_DIRECTORIES']&.split(',') || (ENV['GOOBI_MOUNT'] && [File.join('/', ENV['GOOBI_MOUNT'], 'dcs')]) || ['/brbl-dsu/dcs']
+  end
+
+  def self.system_user
+    system_user = User.find_by_uid('System')
+    unless system_user
+      system_user = User.new(uid: 'System')
+      Rails.logger.error("Unable to save system user") unless system_user.save!
+    end
+    system_user
   end
 end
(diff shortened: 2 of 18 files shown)


app/jobs/mets_directory_scan_job.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+# Very simple job.  Scan needs to occur in a job so that the mounts are available
+class MetsDirectoryScanJob < ApplicationJob
+  queue_as :default
+
+  def default_priority
+    100
```
maxkadel, line 8: I think lower priority happen faster - is having this essentially at the bottom of our priority list intentional?

spec/jobs/mets_directory_scan_job_spec.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
```
maxkadel, line 1: I wish I knew how to write better job specs.... I feel like these could be doing more for us, but I think that's Future Work.

app/lib/mets_directory_scanner.rb
```diff
@@ -0,0 +1,72 @@
...
+      Find.find(directory) do |path|
+        next unless path =~ /.*_mets\.xml$/ && path !~ /.*xslt_result_mets\.xml$/
+        is_done = File.exist?(File.join(File.dirname(path), "#{indicator_file_prefix}.done"))
+        progress_file = File.join(File.dirname(path), "#{indicator_file_prefix}.progress")
+        File.delete(progress_file) if File.exist?(progress_file) && File.mtime(progress_file) < (Time.now.utc - 1.day)
+        is_in_progress = File.exist?(progress_file)
+        begin
+          unless is_done || is_in_progress
+            #  The following will fail if file is not created by this call:
+            IO.sysopen(progress_file, Fcntl::O_WRONLY | Fcntl::O_EXCL | Fcntl::O_CREAT)
```
maxkadel: I'm not familiar with these options - would it be worth it to add a little bit more about them, or a link to documentation? Not required change, just a thought.

spec/lib/mets_directory_scanner_spec.rb
```diff
@@ -0,0 +1,111 @@
+# frozen_string_literal: true
```
maxkadel, line 1: Nice tests

app/lib/mets_directory_scanner.rb
```diff
@@ -0,0 +1,72 @@
+# frozen_string_literal: true
```
maxkadel, line 1: For your consideration - would it make more sense to have this be a concern - something like `DirectoryScannable` and include it in one of our classes, either the mets document or the batch process, maybe?

maxkadel (approved):
I added a couple comments that are just sharing thoughts, but totally happy to have this merged as-is. Looks really good to me.

MaggieZhaoYale (approved):

MaggieZhaoYale:
> Thank @maxkadel. He did the doc...I just merged it!

Thanks @maxkadel Nice doc.

PR #518, opened 2021-02-15
Return nil for viewing_hint and viewing_direction when intranda namespace is not included in mets

martinlovell:
Check for intranda namespace.  Return nil for viewing_hint and viewing_direction if namespace is not included in document.

Co-authored-by: Lakeisha Robinson <lakeisha.robinson@yale.edu>
Co-authored-by: Maggie Zhao <maggie.zhao@yale.edu>
Co-authored-by: Maura Carbone <maura.carbone@yale.edu>
Co-authored-by: Max Kadel <max@curation_experts.com>

### DIFF
--- a/app/models/mets_document.rb
+++ b/app/models/mets_document.rb
@@ -31,12 +31,12 @@ class MetsDocument
   end
 
   def viewing_direction
-    return nil unless @mets.collect_namespaces.include?("intranda:intranda")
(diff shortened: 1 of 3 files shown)


dl-maura (approved):
I did a test of a JSS item on collections test and I got no errors and the item came in!

PR #530, opened 2021-02-24
Update all parent objects after reassociate

martinlovell:
All parents which children moved need to be updated: update child count, Manifest, and PDF.


dl-maura (approved):
This works, we tested this in test

PR #531, opened 2021-02-24
Provide batch wide events and display them

martinlovell:
- Adds method BatchProcess.batch_processing_event to store batch wide IngestEvents that aren't associated with an existing ParentObject or ChildObject
- Display those events on the batch show page.
- Generate batch wide IngestEvents for Reassociate Child batches when parent or child don't exist or order is not valid.  Batch continues to process other rows.
- Generate batch wide IngestEvent for Recreate PTIFF batches when a child oid does not exist.  Batch continues to process other rows.
- Updates batch table to show number of batch wide messages, updates batch show to list the messages
- Update batch form to force user to choose an option (instead of defaulting to Create Parent Objects to prevent errors)

After (batch index/new batch page):
image

After (batch show page):
image

### DIFF
--- a/app/controllers/batch_processes_controller.rb
+++ b/app/controllers/batch_processes_controller.rb
@@ -9,6 +9,8 @@ class BatchProcessesController < ApplicationController
 
   def index
     @batch_process = BatchProcess.new
+    # force user to choose action in form
+    @batch_process.batch_action = nil
 
     respond_to do |format|
       format.html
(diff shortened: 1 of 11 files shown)


MaggieZhaoYale (approved):
This is really nice enhancement. 
Generate batch wide IngestEvents for Reassociate Child batches when parent or child don't exist or order is not valid. Batch continues to process other rows. This is really nice. 
The batch process table looks much much better. 
Thank you.

PR #546, opened 2021-03-02
1121 show user roles for adminset

alishaevn:

# Ticket
https://app.zenhub.com/workspaces/yul-dc-5e50083561915b11fc9e5850/issues/yalelibrary/yul-dc/1121

# Summary
- the admin set "show" page has a table to display all users with the role of "viewer"
- the admin set "show" page has a table to display all users with the role of "editor"

# Demo
| table with users | table without users |
| --- | --- |
| image |  image |

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -31,28 +31,12 @@ class User < ApplicationRecord
     has_role?(:sysadmin)
   end
 
-  def editor=(value)
-    if value.present? && value && value != '0'
-      add_role :editor
-    else
-      remove_role :editor
-    end
+  def editor(admin_set)
+    has_role?(:editor, admin_set)
   end
 
-  def editor
-    has_role?(:editor)
-  end
-
-  def viewer=(value)
-    if value.present? && value && value != '0'
-      add_role :viewer
-    else
-      remove_role :viewer
-    end
-  end
-
-  def viewer
-    has_role?(:viewer)
+  def viewer(admin_set)
+    has_role?(:viewer, admin_set)
   end
 
   def deactivate!
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -31,6 +31,30 @@ class User < ApplicationRecord
     has_role?(:sysadmin)
   end
 
+  def editor=(value)
+    if value.present? && value && value != '0'
+      add_role :editor
+    else
+      remove_role :editor
+    end
+  end
+
+  def editor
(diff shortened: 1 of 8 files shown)


app/models/user.rb
```diff
@@ -31,6 +31,30 @@ def sysadmin
     has_role?(:sysadmin)
   end
 
+  def editor=(value)
```
martinlovell: Should we leave out the global (not tied to admin set) editor and viewer role getter/setters, since we're attaching those roles to admin sets, and having the global role might be confusing?

mikeapp (changes requested):
See comment.  The role relationship should be from the user to an instance of AdminSet, not a global role.

app/models/user.rb
alishaevn: @martinlovell @mikeapp we made the requested changes. please check again! 🙂

app/views/admin_sets/_roles_tables.html.erb
```diff
@@ -0,0 +1,33 @@
+<table class='table table-striped'>
+  <thead class='thead-dark'>
+    <tr>
+      <th scope='col'> Viewers </th>
+    </tr>
+  </thead>
+  <tbody>
+    <% User.all.each do |user| %>
```
mikeapp: Instead of iterating over all of the users you can do this:
`User.with_role(:viewer, @admin_set)`
It runs a single join query to get all of the users who have the specified role.

Looking at it on screen, it would also be useful to sort the list (though this wasn't in the issue):
`User.with_role(:viewer, @admin_set).order('last_name ASC')`

mikeapp (changes requested):
* See inline comment about finding users with the given role using a join instead of by iterating.
* Looking at 1122 draft PR, I noticed that the header can drift down if there are no users listed.  Could you align them to the top?  Example:
Screen Shot 2021-03-03 at 2 35 51 PM

alishaevn:
>     * See inline comment about finding users with the given role using a join instead of by iterating.
> 
>     * Looking at 1122 draft PR, I noticed that the header can drift down if there are no users listed.  Could you align them to the top?  Example:
> 
> 
> Screen Shot 2021-03-03 at 2 35 51 PM

we noticed that on 1122 too, lol. will certainly make sure that's fixed.

mikeapp (approved):

PR #547, opened 2021-03-03
1122 add or remove users in admin set

alishaevn:

# Ticket
https://app.zenhub.com/workspaces/yul-dc-5e50083561915b11fc9e5850/issues/yalelibrary/yul-dc/1122

# Summary
- users can be added as a viewer _or_ editor per admin set
- users can be removed from their role per admin set

# Demo
https://share.getcloudapp.com/4gu1E1PK

# Resources
- https://rubyinrails.com/2019/07/11/rails-routes-member-vs-collection/

### DIFF
--- a/app/controllers/roles_controller.rb
+++ b/app/controllers/roles_controller.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+class RolesController < ApplicationController
+  before_action :set_user
+  before_action :set_item
+
+  def create
+    if @item && @user.has_role?(params[:role], @item)
+      redirect_back(fallback_location: root_path, flash: { alert: 'User already had this role' })
+    elsif @item
+      # each user only gets one role per item, remove all others first.
+      @user.roles.where(resource: @item).destroy_all
+      @user.add_role(params[:role], @item)
+      redirect_back(fallback_location: root_path, notice: "Role Added to User")
+    else
+      @user.add_role(params[:role])
+      redirect_back(fallback_location: root_path, notice: "Role Added to User")
+    end
+  end
+
+  def set_user
+    @user = User.find_by(uid: params[:uid])
+    return true if @user
+
+    redirect_back(fallback_location: root_path, flash: { alert: 'User UID not found' })
+    false
+  end
+
+  def set_item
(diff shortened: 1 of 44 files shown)


mikeapp (commented):
Sorry, looking at the wrong PR - will comment in 1121!

app/controllers/roles_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+  before_action :set_user
+  before_action :set_item, only: [:create]
+
+  def create
+    if @item && @user.has_role?(params[:role], @item)
+      redirect_back(fallback_location: root_path, flash: { alert: "User: #{@user.uid} is already assigned as #{params[:role]}" })
+    elsif @item
+      # each user only gets one role per item, remove all others first.
+      # This removes the role for all users who have that role, not only @user
+      # @user.roles.where(resource: @item).destroy_all
```
dylansalay: This is destroying the role for all users who share that role, not just the user we are adding a new role for.

app/controllers/roles_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+      params[:role] == "viewer" ? @user.remove_role(:editor, @item) : @user.remove_role(:viewer, @item)
+      @user.add_role(params[:role], @item)
+      redirect_back(fallback_location: root_path, notice: show_notice(@user, params[:role]))
+    else
+      @user.add_role(params[:role])
+      redirect_back(fallback_location: root_path, notice: show_notice(@user, params[:role]))
+    end
+  end
+
+  # We tried using the standard destroy action but again, passing in the role id destroys the role for all users who are sharing that role, not just the user whose role we want to destroy
```
dylansalay: The Rails standard destroy action of destroying a Role is the same problem we are seeing above. Users who are editors all share the same editor role. So destroying that role destroys the role for all users.

orangewolf (approved):

app/models/user.rb
```diff
@@ -27,6 +28,10 @@ def sysadmin=(value)
     end
   end
 
+  def find_role(role, admin_set)
```
martinlovell, line 31: This looks problematic and unused. You'd need resource_type to find a unique role.

PR #548, opened 2021-03-03
Create relationship between ParentObject and ChildObject to AdminSet

martinlovell:


app/models/parent_object.rb
```diff
@@ -47,6 +48,7 @@ def self.extent_of_digitizations
   def initialize(attributes = nil)
     super
     self.use_ladybird = true
+    self.admin_set = AdminSet.find_by_key("brbl")
```
mikeapp: This line can be removed.   Sorry if I wasn't clear, the behavior I want is for the migration or seeds file to assign everything that doesn't have an admin set to brbl, which you have done.

mikeapp (changes requested):
See my comment in the initializer.

mikeapp (approved):

PR #564, opened 2021-03-15
Reindex all in batches

martinlovell:
Ordering by OID when taking chunks since it's indexed.
It's possible a parent with an out-of-sequence OID could be ingested after the first job starts and get skipped, but that one is already re-indexed since it was just ingested, so it should be ok.

### DIFF
--- a/app/jobs/solr_reindex_all_job.rb
+++ b/app/jobs/solr_reindex_all_job.rb
@@ -3,21 +3,21 @@
 class SolrReindexAllJob < ApplicationJob
   queue_as :solr_index
 
-  def job_limit
+  def self.job_limit
     5000
   end
 
-  def solr_batch_limit
+  def self.solr_batch_limit
     500
   end
 
   def perform(start_position = 0)
     solr = SolrService.connection
     # Groups of limit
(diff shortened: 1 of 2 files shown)


app/jobs/solr_reindex_all_job.rb
```diff
@@ -3,13 +3,26 @@
...
+
+  def perform(start_position = 0)
     solr = SolrService.connection
-    # Groups of 500
-    ParentObject.find_in_batches do |group|
-      solr.add(group.map(&:to_solr).compact)
-      solr.commit
+    # Groups of limit
+    parent_objects = ParentObject.order(:oid).offset(start_position).limit(SolrReindexAllJob.job_limit)
+    last_job = parent_objects.count < SolrReindexAllJob.job_limit
```
alishaevn, line 18: @martinlovell is there a point that `last_job` would ever return false since we're already limiting `parent_objects` to 5000?

alishaevn: nevermind! figured it out. 🙂

alishaevn (approved):

PR #565, opened 2021-03-17
1137 respect permissions when creating batch jobs

alishaevn:
Co-author: Martin Lovell <martin.lovell@yale.edu>

# Ticket
https://app.zenhub.com/workspaces/yul-dc-5e50083561915b11fc9e5850/issues/yalelibrary/yul-dc/1137

# Expected behavior
- users with an editor role on an admin set can create or update parent objects for that set using the "create parent objects" action
- users without an editor role on an admin set cannot create or update parent objects for that set using the "create parent objects" action
- if a csv has oids with various admin sets, only the sets where the user is an editor will be created/updated
- for failed uploads, there will be an ingest error noted on the batch process detail page

# Demo
https://user-images.githubusercontent.com/29032869/111508394-e6681100-8708-11eb-938a-840234ed52bc.mp4

### DIFF
--- a/app/models/batch_process.rb
+++ b/app/models/batch_process.rb
@@ -104,7 +104,6 @@ class BatchProcess < ApplicationRecord # rubocop:disable Metrics/ClassLength
         fresh = true
       end
       next if fresh
-      next unless user_update_parent_permission(index, po)
 
       po.metadata_update = true
       po.admin_set = admin_set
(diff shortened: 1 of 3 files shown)


app/models/batch_process.rb
```diff
@@ -95,22 +95,16 @@ def create_parent_objects_from_oids(oids, metadata_sources, adminset_keys)
...
+      next unless user_create_permission(index, admin_set, oid)
+
       po = ParentObject.where(oid: oid).first_or_create do |parent_object|
         # Only runs on newly created parent objects
         setup_for_background_jobs(parent_object, metadata_source)
         parent_object.admin_set = admin_set
         fresh = true
-        unless current_ability.can?(:create, parent_object)
-          batch_processing_event("Skipping row [#{index}] because of access violation for parent: #{oid}", 'Skipped Row')
-          parent_object.destroy
```
alishaevn, line 105: we removed this because we were getting a `frozen hash` error, which we believe is related to solr trying to delete the record after it was removed from the db.
https://www.ruby-forum.com/t/cant-modify-frozen-hash-during-validation/147127/4

martinlovell (approved):

app/models/batch_process.rb
```diff
@@ -163,12 +157,22 @@ def configure_parent_object(child_object, parents)
     parents
   end
 
-  def user_update_permission(child_object, parent_object)
+  def user_create_permission(index, admin_set, oid)
+    user = self.user
+    unless user.has_role?(:editor, admin_set)
```
alishaevn, line 162: we're using `has_role` here instead of `current_ability.can?` because we haven't created a parent object yet.

PR #587, opened 2021-04-06
Mark rights statement as html_safe

dylansalay:

# Ticket
https://app.zenhub.com/workspaces/yul-dc-5e50083561915b11fc9e5850/issues/yalelibrary/yul-dc/1103

# Expected behavior
- The Rights field displays HTML links as links
- Only a tags and their href attribute should come through; All other tags are scrubbed. 

# demo
| before | after |
|---|---|
|Screen Shot 2021-04-07 at 9 40 42 AM|Screen Shot 2021-04-07 at 9 40 10 AM|

### DIFF
--- a/app/views/parent_objects/show.html.erb
+++ b/app/views/parent_objects/show.html.erb
@@ -63,7 +63,7 @@
 
 <p>
   <strong>Rights Statement:</strong>
-  <div class='rights-statement'><%= @parent_object.rights_statement.html_safe %></div>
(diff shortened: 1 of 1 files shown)


martinlovell (commented):
You can sanitize if first so that only 'a' tags and href attributes are allowed:
(https://api.rubyonrails.org/classes/ActionView/Helpers/SanitizeHelper.html#method-i-sanitize)

mlooney (approved):

PR #603, opened 2021-05-07
Add logs for manifest actions

edejesusyale:
yalelibrary/YUL-DC#1238

**Story**

For each Parent Object, we need to track changes to its associated IIIF Manifest.   This data will be used to generate an Activity Stream as described in the  IIIF Change Discovery API https://iiif.io/api/discovery/   Essentially this requires a log of CRUD actions on the Manifest, presented in chronological order.

Note: we'll eventually decide where/how to generate the ActivityStreams, but that's for another ticket.  This may involve creating a database table to log this data, or sending it to another system.

**Acceptance**
-  Create a log entry for each of the events below.  
   - [x] `Create` when a Manifest is first published, or published from a deleted or non-accessible  (i.e., not Public or YCO)  state
   - [x] `Update` when a Manifest is updated
      - [x] Changes that do not change the contents of the Manifest should not create an Update event.   (Store manifest checksum on the parent object and use that for comparison.)
   - [x] `Delete` when a Manifest is deleted or the Parent is made inaccessible (i.e., not Public or YCO)
   - [x] A timestamp is logged for each event.

### DIFF
--- a/app/jobs/generate_manifest_job.rb
+++ b/app/jobs/generate_manifest_job.rb
@@ -11,6 +11,8 @@ class GenerateManifestJob < ApplicationJob
     parent_object.current_batch_process = current_batch_process
     parent_object.current_batch_connection = current_batch_connection
     generate_manifest(parent_object)
+    update_manifest_log(parent_object)
+
     index_to_solr(parent_object)
     GeneratePdfJob.perform_later(parent_object, current_batch_process, current_batch_connection)
   end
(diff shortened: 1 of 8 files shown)


app/jobs/generate_manifest_job.rb
```diff
@@ -34,6 +36,19 @@ def generate_manifest(parent_object)
...
   end
 
+  def update_manifest_log(parent_object)
+    ## logs create if checksum is nil
+    if parent_object.create_manifest_log?
+      parent_object.create_manifest_log
+    else
+      # #checks if checksum is different before update
+      updated_checksum = Digest::SHA1.hexdigest parent_object.iiif_manifest.to_yaml
+      parent_object.manifest_logs << { status: "Update", timestamp: Time.zone.now } unless updated_checksum.eql? parent_object.manifest_checksum
```
martinlovell: Should parent_object.manifest_checksum get updated right after this line?

app/models/parent_object.rb
```diff
@@ -283,6 +286,25 @@ def iiif_manifest
...
+
+  def delete_manifest_log
+    manifest_logs << { status: "Delete", timestamp: Time.zone.now }
+    self.manifest_checksum = nil
+  end
+
+  def create_manifest_log
+    self.manifest_logs ||= []
+    self.manifest_logs << { status: "Create", timestamp: Time.zone.now }
+    self.manifest_checksum = Digest::SHA1.hexdigest iiif_manifest.to_yaml
```
martinlovell: This and generate_manifest_job.rb:48 could be moved to a new `IiifPresentation.checksum` method.

Should it checksum `iiif_manifest.to_json` instead of `to_yaml`, so it matches the checksum of the file on S3? (`to_json(pretty: true)` so they match up with `IiifPresentation.save`.)

martinlovell (approved):

PR #649, opened 2021-06-03
Styled global footer

FCRodriguez7:
**Story:** 

Create a global footer for versioning info, so that it will be clear that this information applies to the whole app and not to whatever is on the current page. (See attached design mockup for reference)

Acceptance: 

- [ ] Create a global footer with left-aligned versioning info (same as the current app, but moved into a specific region on the page)

----
**Actual Modification**
image


MaggieZhaoYale (approved):
Looks good

PR #679, opened 2021-06-25
1384 store dependent objects on fetch

martinlovell:
Maintains dependent objects on metadata fetch.
Removes ParentObject.complete_fetch() which is not used/needed.


martinlovell:
I disabled one test and edited activity_stream_reader to get tests passing, etc.  activity_stream_reader and related tests will be updated in #678

MaggieZhaoYale (approved):

PR #686, opened 2021-06-29
Store the state of datatables and populate filters

martinlovell:
- Stores the state of the datatable using the built in "stateSave" option for datatables.
- Fills the filter boxes/selects with the filters being applied when the state is restored by the datatable.
- Adds a "Clear Filters" button so that users can clear the filters.  (This is especially important because if there is a hidden column with a filter, it could be difficult to clear, and refreshing the page will no longer clear filters.)

FilterRetain


MaggieZhaoYale (approved):
looks good!

PR #688, opened 2021-06-29
1395-export-child-oid-sort-order

alishaevn:
# Ticket
https://app.zenhub.com/workspaces/yul-dc-5e50083561915b11fc9e5850/issues/yalelibrary/yul-dc/1395

# Expected behavior
- after running an "export child oids" batch process job, the csv that is created will have headers in the following order:
  - parent_oid
  - child_oid
  - order
  - parent_title
  - label
  - caption
  - viewing_hint
- the parent oids will be sorted first
- then, each child will be sorted in order according to its parent grouping

# Demo
image

### DIFF
--- a/app/models/concerns/csv_exportable.rb
+++ b/app/models/concerns/csv_exportable.rb
@@ -4,20 +4,19 @@ module CsvExportable
   extend ActiveSupport::Concern
 
   def headers
-    ["parent_oid", "child_oid", "order", "parent_title", "label", "caption", "viewing_hint"]
+    ['parent_oid', 'child_oid', 'order', 'parent_title', 'label', 'caption', 'viewing_hint']
   end
 
   def output_csv
-    return nil unless batch_action == "export child oids"
+    return nil unless batch_action == 'export child oids'
+
     CSV.generate do |csv|
       csv << headers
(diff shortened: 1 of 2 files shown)


martinlovell:
I think the issue was for the order of the rows in the csv, not the order of the columns.
So, it should sort rows in the csv first by parent oid, then by child order.

mikeapp:
Correct, the sort order is the issue.

alishaevn:
@martinlovell @mikeapp ahhh. ok.

alishaevn:
@martinlovell  @mikeapp
- are the column headings still supposed to be parent oid, child oid, then order?

Can you confirm the below?
- the parent oid column should be sorted first so that all parent oids are grouped together
- the child oids are then sorted according to their order number

e.g.

| parent oid | child oid | order|
| --- | --- | --- |
| 2005512 | 1030368 | 1 |
| 2005512 | 1032318 | 2 |
| 16414889 | 16424346 | 1 |
| 16414889 | 16424347 | 2 |
| 16414889 | 16424348 | 3 |

mikeapp:
Correct, just like that.

jpengst (approved):

PR #692, opened 2021-07-01
1413 Map new metadata fields from Metadata Cloud

FCRodriguez7:
**Story**

A number of new metadata fields are being added to Metadata Cloud in order to support the Archival Context work.

Added a special fieldType: "repositoryName" which gets the repository display_string from a cache.  Mapped to "Repository" property.

We added "ArchivalSort" as a special property (not in the json mapping file).  We implemented it based on your examples and sample code.

We updated the AncestorTitles property by appending the call number into the resource title and adding the repository.
We also added the repository URI to the AncestorIdentifiers array so that it corresponds to the AncestorTitles.

We added a new field called "AncestorDisplayStrings" for this list.  It is similar to AncestorTitles but favors display_string over title.  It looks like it is getting populated with the series macro and some other macros in some cases. 
We also added the repository to this list so the indexes correspond with AncestorTitles and AncestorIdenifiers.

**Acceptance**
- [x] index `repository` from Metadata Cloud as  `repository_ssi`
- [x] index `archivalSort` as `archivalSort_ssi`
- [x] index `ancestorTitles` as `ancestorTitles_tesim`
- [x] index the penultimate, second to last, value in the `ancestorTitles` array (if present) as `collection_title_ssi`
- [x] index `ancestorDisplayStrings` as `ancestorDisplayStrings_tesim`

----
Testing Notes

One can use aspace 2005512 to test this code modification. 

-----
**Demo** 

Solr Archivel Feature


MaggieZhaoYale (approved):
looks good

PR #703, opened 2021-07-08
1416 Add "Collection Note" field to IIIF Manifest

FCRodriguez7:
Co-Authored-By: Dr Eric DeJesus <eric.dejesus@yale.edu>

**Story**

The IIIF Manifest metadata should indicate the collection that contains the object.

**Acceptance**
- [ ] Add a field to the Manifest
  - Label is "Collection Note"
  - Value is the same as #1414.  It should be text only, not a link
  - See the mockup for position within the metadata field list.  **Note that the collection title/date fields will also need to be added, so position under Call Number for now.**


FCRodriguez7:
image

PR #709, opened 2021-07-13
1416 modify collection note field

FCRodriguez7:
**Story**

The IIIF Manifest metadata should indicate the collection that contains the object.

**Acceptance**
- [x]  Reverse the array
- [x] Add the sourceNote mapping field for Ladybird objects.


MaggieZhaoYale (approved):
The change is cleaner and clearer. Very nice!

PR #710, opened 2021-07-13
Adds delete batch job 

jpengst:
# Summary
Create batch job to delete parent objects by uploading a csv of oids.

# Related Ticket
[#1407](https://app.zenhub.com/workspaces/yul-dc-5e50083561915b11fc9e5850/issues/yalelibrary/yul-dc/1407)

# Video

https://user-images.githubusercontent.com/36549923/127907834-242fdfc3-3716-4aac-aba1-7490488eb320.mp4


app/controllers/child_objects_controller.rb
```diff
@@ -70,7 +70,14 @@ def destroy
...
     def set_child_object
-      @child_object = ChildObject.find(params[:id]) if params[:id]
+      begin
+        @child_object = ChildObject.find(params[:id]) if params[:id]
+      rescue ActiveRecord::RecordNotFound
+        respond_to do |format|
+          format.html { redirect_to parent_objects_url, notice: "Child object, oid: #{params[:id]}, was not found in local instance." }
+          format.json { head :no_content }
+        end
+      end
```
orangewolf: No real reason to raise an exception here. 

```suggestion
      @child_object = ChildObject.find_by(id: params[:id]) if params[:id]
      if @child_object.blank?
        respond_to do |format|
          format.html { redirect_to parent_objects_url, notice: "Child object, oid: #{params[:id]}, was not found in local instance." }
          format.json { head :no_content }
        end
      end
```

app/controllers/parent_objects_controller.rb
```diff
@@ -141,7 +141,14 @@ def invalidate_admin_set_edit
 
     # Use callbacks to share common setup or constraints between actions.
     def set_parent_object
-      @parent_object = ParentObject.find(params[:id])
+      begin
+        @parent_object = ParentObject.find(params[:id])
```
orangewolf: see above

app/jobs/remove_from_metadata_cloud_csv_job.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class RemoveFromMetadataCloudCsvJob < ApplicationJob
+  queue_as :default
+
+  def default_priority
+    -50
```
orangewolf: this is a very hi propriety, is that what you mean to do here?

app/models/batch_process.rb
```diff
@@ -106,6 +106,33 @@ def create_parent_objects_from_oids(oids, metadata_sources, adminset_keys)
       end
     end
   end
+
+  def delete_parent_object(oids, metadata_sources, adminset_keys)
```
orangewolf: use admin_set_key, just to be consistent please

app/models/batch_process.rb
```diff
@@ -106,6 +106,33 @@ def create_parent_objects_from_oids(oids, metadata_sources, adminset_keys)
...
+    admin_set_hash = {}
+    oids.zip(metadata_sources, adminset_keys).each_with_index do |record, index|
+      oid, metadata_source, adminset_key = record
+      admin_set = admin_set_hash[adminset_key]
```
orangewolf: all this admin_set questing belongs in its own method

app/models/batch_process.rb
```diff
@@ -106,6 +106,33 @@ def create_parent_objects_from_oids(oids, metadata_sources, adminset_keys)
...
+        admin_set = AdminSet.find_by_key(adminset_key)
+        admin_set_hash[adminset_key] = admin_set
+      end
+      if admin_set.nil?
+        batch_processing_event("Skipping row [#{index + 2}] with unknown admin set [#{adminset_key}] for parent: #{oid}", 'Skipped Row')
+        next
+      end
+      next unless user_edit_permission(index, admin_set, oid)
+
+      if ParentObject.where(oid: oid).count.negative?
```
orangewolf: what does `ParentPbject.destroy` do? is it different from `po = ParentObject.find_by(oid: oid); po.destroy if po.present?`

app/models/batch_process.rb
```diff
@@ -121,6 +148,12 @@ def refresh_metadata_cloud_csv
     create_parent_objects_from_oids(oids, metadata_sources, admin_sets)
   end
 
+  def remove_from_metadata_cloud_csv
+    metadata_sources = parsed_csv.entries.map { |r| r['source'] }
```
orangewolf: its expensive to run through these columns twice. can we do a single pass instead?

orangewolf (changes requested):

app/jobs/remove_from_metadata_cloud_csv_job.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class RemoveFromMetadataCloudCsvJob < ApplicationJob
```
mikeapp: Can we drop the MetadataCloud reference here?  It's a misnomer in the other job as well.  Something like "DeleteObjectsJob" would be fine.

jpengst: Yep, no problem! Thanks for the feedback

app/controllers/child_objects_controller.rb
K8Sewell: We attempted to use the `find_by` method but that resulted in the error below.  So we stuck with what we had.  Open to trying again though.

image

app/controllers/parent_objects_controller.rb
K8Sewell: same as above

app/models/batch_process.rb
```diff
@@ -84,45 +84,68 @@ def created_file_name
...
-    create_parent_objects_from_oids(oids, metadata_sources, admin_sets)
+    parsed_csv.each_with_index do |row, index|
+      oid = row['oid']
+      metadata_source = row['source']
+      admin_set = editable_admin_set(row['admin_set'], oid, index)
+      next unless admin_set
+
+      parent_object = ParentObject.find_or_initialize_by(oid: oid)
+      # Only runs on newly created parent objects
+      next unless parent_object.new_record?
```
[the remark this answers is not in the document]
K8Sewell: Just updated this and should be through the pipeline soon.  Everything passed locally.

martinlovell (approved):

orangewolf (approved):

PR #742, opened 2021-08-11
Adds update parent batch job

K8Sewell:
# Summary
Adds a batch process for updating certain fields on parent objects with a csv upload.

# Related Ticket
[#1466](https://app.zenhub.com/workspaces/yul-dc-5e50083561915b11fc9e5850/issues/yalelibrary/yul-dc/1466)

# Video
https://user-images.githubusercontent.com/36549923/129114415-ba73dbec-d31b-4113-b1ca-d916a9536b24.mp4


app/models/concerns/updatable.rb
```diff
@@ -0,0 +1,71 @@
...
+      processed_fields[f.to_sym] = (
+        row[f].present? &&
+        row[f] != parent_object.send(f)
+      ) ? row[f] : parent_object.send(f)
+    end
+    validation_fields.each do |k, v|
+      processed_fields[k.to_sym] = (
+        row[k].present? &&
+        row[k] != parent_object.send(k) &&
+        (ParentObject.send(v).include? row[k])
```
martinlovell: You should probably add a batch message if the validation fails for one of the fields.

mikeapp: We definitely need to let the user know so they can correct the problem and resubmit for that OID.

jpengst: Image 2021-08-12 at 4 28 04 PM

K8Sewell: We've added some detailed batch messages.  And are just cleaning up the specs now.

image

martinlovell (approved):

app/models/concerns/updatable.rb
martinlovell: Nice messages!

PR #754, opened 2021-08-23
Reassociate child column values

jpengst:
# Summary
Re-associating Child OIDS batch process now allows 
- partial columns in csvs
- ability to set nil values for label, caption, and viewing_hint

# Related Ticket
[#1546](https://app.zenhub.com/workspaces/yul-dc-5e50083561915b11fc9e5850/issues/yalelibrary/yul-dc/1546)

# Screenshots
Before Reassociate Child OIDs:  
Image 2021-08-23 at 2 54 18 PM  
  
After Reassociate Child OIDs:  
Image 2021-08-23 at 2 55 50 PM  
  
Before and After in Rails Console
Image 2021-08-23 at 2 56 10 PM

Error message provided when using invalid controlled vocabulary for viewing hint
image

### DIFF
--- a/app/models/concerns/reassociatable.rb
+++ b/app/models/concerns/reassociatable.rb
@@ -18,14 +18,19 @@ module Reassociatable
       po = load_parent(index, row["parent_oid"].to_i)
       next unless co.present? && po.present?
 
+      # byebug
+
+      # headers - check if headers present before update
       attach_item(po)
       attach_item(co)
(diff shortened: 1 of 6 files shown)


FCRodriguez7 (approved):

martinlovell (commented):
It doesn't seem like this checks that the column exists in the CSV.
(See issue: If entire column is missing, do not update that property for any object)

PR #756, opened 2021-08-30
Changed fulltext to allow partial text status

FCRodriguez7:
Co-authored-by: Eric DeJesus <eric.dejesus@yale.edu>

**This should be done in a new branch, not `fulltext`**
**The new branch name is next_phase_fulltext_feature**

**Story**

Objects may be partially transcribed, or reassociations may result in the combination of a transcribed object with one that has no transcription.  This should be indicated in the Full Text search facet.   

The current approach of checking the first page for transcription and then assuming that the entire object is transcribed will no longer work.  To avoid performing repeated scans of AWS S3, we will record the full-text status at the child level on ingest. 

**Acceptance**
- [x] Add a `full_text` field to the Child Object model with type Boolean, defaulting to false
- [x] The Parent Object will need to be refactored to be able to indicate that it has partial full text (i.e. some of its children have `full_text` true and others false).  Currently there's a method to return a boolean (`full_text?`).   All values should be based on the values set in the Child Objects, not a real-time scan of S3.
- [x] If an object has partial full text, in `solr_indexable.rb` index the value "Partial" (currently the value is "Yes" or "No"). 
- [x] the logic ought to account for when the object has partial text

### DIFF
--- a/app/models/batch_process.rb
+++ b/app/models/batch_process.rb
@@ -89,10 +89,10 @@ class BatchProcess < ApplicationRecord # rubocop:disable Metrics/ClassLength
     object.authoritative_metadata_source = MetadataSource.find_by(metadata_cloud_name: (metadata_source.presence || 'ladybird')) if object.class == ParentObject
     object.current_batch_process = self
     object.current_batch_connection = batch_connections.build(connectable: object)
-    if object.class == ChildObject
-      object.full_text = object.remote_ocr
-      object.current_batch_connection.save!
-    end
+    return unless object.class == ChildObject
+
+    object.full_text = object.remote_ocr
+    object.current_batch_connection.save!
   end
 
   def editable_admin_set(admin_set_key, oid, index)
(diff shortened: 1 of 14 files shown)


martinlovell (changes requested):
I really like the new undelayed: true spec helpers.

I think that ChildObject is supposed to get a new field in the DB for full_text, which will get set in 1487.  So, I think that first checkbox on issue 1481 needs a migration.

martinlovell (commented):
- For the solr record, "fulltext_tesim" should have values of Yes, No, or Partial. That way the facet will work with blacklight.  (I'm not sure if any other solr fields are needed.)

- Is it necessary to update the children in ParentObject.create_child_records? (lines 100-103)  It seems like the upsert should set them.  Editing and saving each child could be slow, but the upsert is quick.

- Also, it seems  a little risky the way ParentObject.partial_fulltext? is setup since you might get a different value based on whether you called full_text? first.  Maybe there's a way to take advantage of only going through the children once for both, but still calculating partial_fulltext? if full_text? wasn't called first, instead fo defaulting to false.

martinlovell (approved):

PR #762, opened 2021-09-02
Adds rake task to generate labels for child objects

K8Sewell:
# Summary
Adds a rake task to generate pagination labels for child objects if the label is empty.

# Related Ticket
[#1528](https://app.zenhub.com/workspaces/yul-dc-5e50083561915b11fc9e5850/issues/yalelibrary/yul-dc/1528)

# Screenshot
image


mikeapp:
This should target the list of specific OIDs attached to the main issue, not all parents.  Many labels in Beinecke are intentionally blank.  

It is only the JSS collection that has this problem, and basically a workaround until we upgrade UV.

FCRodriguez7 (commented):
Could we add code coverage to this?

FCRodriguez7 (approved):

PR #768, opened 2021-09-10
Prevent multiple solr index jobs to be queued simultaneously

martinlovell:
- Only queue a SolrIndexJob if there is not one already in place for the parent
- Change Manifest job to perform index asynchronously using a separate job (as we do in other places)
- Move processing events that were in manifest jobs synchronous indexing to parent object so they are always sent to the batch


martinlovell:
Closing because #769 (which removes the update hook) is closer to discussed approach.

PR #809, opened 2021-10-12
Ensures project identifier can index to solr document

K8Sewell:
# Summary
Ensures that a value will be saved to the solr document for the project identifier field.

# Related Ticket
[#1655](https://app.zenhub.com/workspaces/yul-dc-5e50083561915b11fc9e5850/issues/yalelibrary/yul-dc/1665)

# Screenshots
image

image

image

### DIFF
--- a/app/models/concerns/solr_indexable.rb
+++ b/app/models/concerns/solr_indexable.rb
@@ -106,7 +106,7 @@ module SolrIndexable
       orbisBarcode_ssi: barcode,
       orbisBibId_ssi: bib, # may change to orbisBibId
       preferredCitation_tesim: json_to_index["preferredCitation"],
-      project_identifier_tesi: json_to_index["project_identifier"],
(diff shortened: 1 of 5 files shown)


app/models/concerns/solr_indexable.rb
```diff
@@ -257,4 +257,9 @@ def ancestor_structure(ancestor_title)
   def generate_hash
     Digest::MD5.hexdigest oid.to_s
   end
+
+  def generate_pid(project_identifier, oid)
+    parent_object = ParentObject.find_by(oid: oid)
```
martinlovell: Is parent_object the same as "self" here?  If you aren't explicitly reloading it, you could just access self&.project_identifier instead of loading the parent_object with find_by.

K8Sewell: self did work.  Thanks for helping me avoid a db call!

FCRodriguez7 (approved):

PR #811, opened 2021-10-12
added digitization note to solr index

jpengst:
# Summary

In order to display "Digitization Note" in Blacklight, digitization_note needs to be indexed by solr.


app/models/concerns/solr_indexable.rb
```diff
@@ -257,4 +257,9 @@ def ancestor_structure(ancestor_title)
   def generate_hash
     Digest::MD5.hexdigest oid.to_s
   end
+
+  def generate_digitization_note(digitization_note, oid)
+    parent_object = ParentObject.find_by(oid: oid)
```
martinlovell, line 262: I think parent_object is self  (SolrIndexable is part of ParentObject), so it doesn't need to be loaded from the db.

FCRodriguez7 (approved):

PR #829, opened 2021-10-28
Partial full text implementation

martinlovell:
Merges changes from Partial Full Text implementation into Main.


db/schema.rb
```diff
@@ -74,6 +74,7 @@
     t.datetime "ptiff_conversion_at"
     t.string "mets_access_master_path"
     t.integer "original_oid"
+    t.boolean "full_text", default: false
```
FCRodriguez7, line 76: This line was below, wondering why the line moved? And how does the move impact the overall file?

martinlovell: Thanks for catching that.
That was how it got generated on my system, but I think it was off.

I recreated if from a fresh DB and committed again:
I did a `cam down` to clear the DB, and then `cam up management` to get a fresh DB based on the migration order in this branch.
Then ran `bundle exec rake db:schema:dump` to re-generate the schema.rb from the DB.
It did change the order, so I committed the new cleaner copy.

FCRodriguez7 (approved):

PR #830, opened 2021-10-29
1707 parent object datatable title style

jpengst:
Title-Style capitalization applied to the Parent Object Datatable:  
Image 2021-10-29 at 12 31 39 PM


jpengst:
I added Title-Style to some others column headers on Child Object, Users, and Batch Process datatables to make them all consistent.

K8Sewell (approved):

PR #836, opened 2021-11-03
Set Solr Heap for Solr for circle ci

martinlovell:


martinlovell:
Closing because more config changes need to happen to get this working in CI.  See #837

PR #845, opened 2021-11-10
Add alternative dashboard for DelayedJob

martinlovell:
Co-authored-by: McClain Looney <mcclain@curationexperts.com>

Creates an alternative dashboard for delayed jobs that uses more reasonable queries to get the statistics about jobs.
The old delayed job dashboard is still available for those brave enough to try it, and when there are not > 500,000 jobs.


MaggieZhaoYale (approved):
https://collections-demo.library.yale.edu/management looks good!

PR #853, opened 2021-11-30
datatable export all

jpengst:
Users can now export all filtered records to a CSV instead of just what's displayed on the table.


mikeapp:
Sorry, didn't realize this was in progress so I rewrote the ticket.  If you have a datatables-based approach that's working then let's stick with that for now.

FCRodriguez7 (approved):

FCRodriguez7 (approved):

PR #866, opened 2022-01-04
Save exported child csv to S3

K8Sewell:
# Summary
To prevent recreating CSV documents upon download the exported CSV will be saved to S3 and then a url to that asset will be served from the batch process show page.

# Related Ticket
[#1814](https://app.zenhub.com/workspaces/yul-dc-5e50083561915b11fc9e5850/issues/yalelibrary/yul-dc/1814)

# Screenshot
image


app/controllers/batch_processes_controller.rb
```diff
@@ -63,17 +63,18 @@ def download_template
     end
   end
 
+  # TODO: get rid of this method...potentially
```
martinlovell, line 66: I agree.

martinlovell (approved):

PR #888, opened 2022-01-20
Adds function to disable button over record threshold

K8Sewell:
# Summary
If the datatable has over 12,000 records the 'Export Matching Entries' will be disabled and will display a user message on hover.  This PR also removes a custom alert that has broken the export in production.

# Related Ticket
[#1800](https://app.zenhub.com/workspaces/yul-dc-5e50083561915b11fc9e5850/issues/yalelibrary/yul-dc/1800)

# Screenshots
### Disabled
Screen Shot 2022-01-20 at 1 49 20 PM

### User Messaging - On Hover
Screen Shot 2022-01-20 at 1 49 04 PM

### DIFF
--- a/app/javascript/packs/application.js
+++ b/app/javascript/packs/application.js
@@ -334,7 +334,7 @@ $( document ).on('turbolinks:load', function() {
   table.clear()
   table.ajax.url($('.is-datatable').data('source')).load(function() {
 
-    var info = $('#child-objects-datatable_info').text().split(' ')
+    var info = $('#child-objects-datatable_info, #parent-objects-datatable_info, #redirected-parent-objects-datatable_info, #batch-processes-datatable_info, #preservica-ingests-datatable_info').text().split(' ')
     var entries = info[5]
     if ( entries > 12000 ) {
       // Disable the button
--- a/app/javascript/packs/application.js
+++ b/app/javascript/packs/application.js
@@ -191,9 +191,6 @@ $( document ).on('turbolinks:load', function() {
           action: newExportAction,
           text: "All Matching Entries",
           className: "export-all"
-          // customize: function (csv) {
-          //   return check_csv(csv)
-          // }
         }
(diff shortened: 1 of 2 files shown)


app/javascript/packs/application.js
```diff
@@ -328,3 +326,29 @@ $( document ).on('turbolinks:load', function() {
...
 })
+
+// This will disable the export all button when there are more than 12K records
+$( document ).on('turbolinks:load', function() {
+  // Get count of records
+  var table = $('.is-datatable').DataTable();
+  table.clear()
+  table.ajax.url($('.is-datatable').data('source')).load(function() {
+
+    var info = $('#child-objects-datatable_info').text().split(' ')
```
martinlovell: Will this have a different ID for parent objects?

K8Sewell: Nope, all of the datatables have the 'is-datatable' class.  For example [here](https://github.com/yalelibrary/yul-dc-management/blob/main/app/views/parent_objects/index.html.erb#L25) on the parent object.  We should be safe with this identifier.

K8Sewell: I realize now you were actually referring to the 'child-objects-datatable_info' identifier.  This will need a refactor.  Thank you for catching this.

martinlovell: I tried putting in an event listener to get the details, and it seems like it's working and updates properly when filtering has changed.
If you could use that instead of parsing the display it may work better, and will cover all tables.
(fnRecordsDisplay is the total with filters applied.)

```
dataTable = $('.is-datatable').dataTable({
 ...
      "fnDrawCallback": function( oSettings ) {
        console.log(oSettings.fnRecordsTotal() + "  -  " + oSettings.fnRecordsDisplay());
      }
 ...
 }
```

mikeapp (approved):

PR #891, opened 2022-01-21
Added basic preservica client

martinlovell:
### Preservica Client with some sample Models

- Adds preservica http client and some models to wrap the XML data
- Models hold preservica client with correct credentials so that data can be loaded.
- Uses "lazy loading" to load related models or properties when they are accessed.

### Requirements

Client expects environment variables:
- PRESERVICA_HOST to be the hostname of the preservica API server.
- PRESERVICA_CREDENTIALS to be JSON in the following format, using admin set keys: 
```json
{"brbl": {"username":"xxxxx", "password":"xxxxx"}} 
```

### Structure

Client is in `app/lib/preservica_client.rb`
Models are in `app/models/preservica`
Fixtures are in `spec/fixtures/preservica`
Tests are in `spec/models/preservica`

### Usage

Traverse the object model, loading children and XML properties as needed:
```
structured_object = StructuralObject.where(admin_set_key: 'brbl', id: "7fe35e8c-c21a-444a-a2e2-e3c926b519c4")
information_objects = structured_object.information_objects
representations = information_objects[0].representations # looking at first information object
content_objects = representations[0].content_objects
generations = content_objects[0].active_generations
bitstreams = generations[0].bitstreams

checksum = bitstreams[0].sha512_checksum
size = bitstreams[0].size

data = bitstreams[0].bits
downloaded_sha = Digest::SHA512.hexdigest data


data.size == size
downloaded_sha == checksum
```


app/models/preservica/bitstream.rb
```diff
@@ -0,0 +1,30 @@
...
+  include PreservicaObject
+
+  def initialize(preservica_client, content_id, generation_id, id)
+    @preservica_client = preservica_client
+    @id = id
+    @content_id = content_id
+    @generation_id = generation_id
+  end
+
+  def checksum
```
mikeapp: There's the potential for multiple checksum algorithms, which will break this.  Please rename to sha512_checksum and retrieve based on the `<xip:FixityAlgorithmRef>`

app/models/preservica/content_object.rb
```diff
@@ -0,0 +1,29 @@
...
+    preservica_client = options[:preservica_client] || PreservicaClient.new(options[:admin_set_key])
+    ContentObjecct.new(preservica_client, options[:id])
+  end
+
+  def initialize(preservica_client, id)
+    @preservica_client = preservica_client
+    @id = id
+  end
+
+  def generations
```
mikeapp: Is this loading only the active generation?  If so, please rename to `active_generation`

mikeapp (changes requested):

app/models/preservica/content_object.rb
martinlovell: changes name to active_generations

config/application.rb
```diff
@@ -39,5 +39,6 @@ class Application < Rails::Application
 
     # Don't generate system test files.
     config.generators.system_tests = nil
+    config.autoload_paths += Dir[Rails.root.join('app', 'models', '**')]
```
orangewolf: app/models should be in autoload paths by default. I think this is redundant?

martinlovell: It seems like it didn't pick up the sub-directories unless it's added with **.

app/lib/preservica_client.rb
```diff
@@ -0,0 +1,114 @@
...
+# Uses environment variables: PRESERVICA_HOST and PRESERVICA_CREDENTIALS
+#
+# rubocop:disable Metrics/ClassLength
+class PreservicaClient
+  def initialize(args)
+    @host = args[:base_url] || "https://#{ENV['PRESERVICA_HOST']}"
+    if args[:admin_set_key]
+      credentials = JSON.parse(ENV['PRESERVICA_CREDENTIALS'])[args[:admin_set_key]]
+      @username = credentials['username']
+      @password = credentials['password']
```
mikeapp, line 14: Suggestion: customize `inspect` on the `PreservicaClient` so that it does not output `@password`.    Since the various model objects have a reference to the client it's likely this will end up in the logs.  Same possibly for the `@login_info` hash, if the refresh token is not short-lived.

mikeapp (changes requested):

app/lib/preservica_client.rb
martinlovell: I updated the PR with these changes.

config/application.rb
mikeapp: PR to add namespaces to this branch: #900

mikeapp (approved):

PR #907, opened 2022-01-31
WIP: Demonstrate issue with Solr save with no children

martinlovell:
This is just to demonstrate the issue with parents not updating solr if they have no children.

This can be demonstrated using the UI by following these steps:
- Choose a test parent that is public, indexed, and searchable in SOLR.
- In Management, reassociate all the children for that parent to some other parent(s). (It's easiest to re-associate with multiple parents so it doesn't end up being redirected.) (At this point, it will no longer be able to update solr).
- In Management, set the parent to "Private" and save.
- Check the Solr record and blacklight: Record will still be in solr from before it was set to private and children were removed.  (Blacklight thinks it's public, so it will serve it up and the IIIF and manifest, etc.)

### DIFF
--- a/spec/models/parent_object_solr_spec.rb
+++ b/spec/models/parent_object_solr_spec.rb
@@ -112,14 +112,33 @@ RSpec.describe ParentObject, type: :model, prep_metadata_sources: true, solr: tr
       parent_object
     end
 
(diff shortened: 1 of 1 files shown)


martinlovell:
Just a demo test...so closing.

PR #912, opened 2022-02-03
added checksum to child object

jpengst:
Added sha512_checksum field to child objects.

### DIFF
--- a/app/controllers/child_objects_controller.rb
+++ b/app/controllers/child_objects_controller.rb
@@ -81,6 +81,6 @@ class ChildObjectsController < ApplicationController
     # Only allow a list of trusted parameters through.
     def child_object_params
       params.require(:child_object).permit(:oid, :caption, :label, :width, :height, :order, :viewing_hint,
-                                           :parent_object_oid, :preservica_content_object_uri, :preservica_generation_uri, :preservica_bitstream_uri)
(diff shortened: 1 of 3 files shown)


app/controllers/child_objects_controller.rb
```diff
@@ -81,6 +81,6 @@ def set_child_object
     # Only allow a list of trusted parameters through.
     def child_object_params
       params.require(:child_object).permit(:oid, :caption, :label, :width, :height, :order, :viewing_hint,
-                                           :parent_object_oid, :preservica_content_object_uri, :preservica_generation_uri, :preservica_bitstream_uri)
+                                           :parent_object_oid, :preservica_content_object_uri, :preservica_generation_uri, :preservica_bitstream_uri, :sha512_checksum)
```
martinlovell, line 84: Should it not be permitted here, so that clients can't set the value?

FCRodriguez7 (approved):

app/controllers/child_objects_controller.rb
FCRodriguez7: I would believe so.

jpengst: Yes, good point! Thanks! I've pushed up a new PR to remove.

PR #954, opened 2022-03-15
Changed the image size in the sitemap and og

FCRodriguez7:
Co-authored-by: Kait Sewell <kait@notch8.com>
Co-authored-by: Amanda Frost <amanda@scientist.com>
Co-authored-by: JP Engstrom <jp@notch8.com>

**Story**

Our sitemap files link to the thumbnail image and the size is `!200,200`.   We should update this to match the image size we going to use for OpenGraph.  See #1249.

**Acceptance**
- [ ] Change sitemap XML image references to `!1200x630`

Ticket 1249
**Acceptance**
- [ ] Change `!200x200` to  `!1200x630` size in the IIIF thumbnail URL

### DIFF
--- a/app/models/child_object.rb
+++ b/app/models/child_object.rb
@@ -128,7 +128,7 @@ class ChildObject < ApplicationRecord
   end
 
   def thumbnail_url
-    "#{IiifPresentationV3.new(parent_object).image_service_url(oid)}/full/!200,200/0/default.jpg"
+    "#{IiifPresentationV3.new(parent_object).image_service_url(oid)}/full/!1200x630/0/default.jpg"
   end
 
   def width_and_height(size)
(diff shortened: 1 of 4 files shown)


app/models/child_object.rb
```diff
@@ -128,7 +128,7 @@ def pyramidal_tiff
   end
 
   def thumbnail_url
-    "#{IiifPresentationV3.new(parent_object).image_service_url(oid)}/full/!200,200/0/default.jpg"
+    "#{IiifPresentationV3.new(parent_object).image_service_url(oid)}/full/!1200x630/0/default.jpg"
```
K8Sewell: Think these may need to be commas not xs

FCRodriguez7: I fixed it.  Copy, cut and dump effect

mikeapp:
I think we can discard this PR?

martinlovell:
Closing since this will be done in blacklight.

PR #958, opened 2022-03-23
Only create PDFs if md5 of Json changes

martinlovell:


spec/jobs/generate_pdf_job_spec.rb
```diff
@@ -53,5 +53,18 @@ def queue_adapter_for_test
...
+      before do
+        stub_request(:head, "https://not-a-real-bucket.s3.amazonaws.com/pdfs/19/16/71/24/19/16712419.pdf")
+            .to_return(status: 200, headers: { "x-amz-meta-pdfchecksum": parent_object_with_authoritative_json.pdf_json_checksum })
+      end
+      it "generate_pdf returns false because it does not generate pdf" do
+        expect(parent_object_with_authoritative_json.generate_pdf).to be_falsey
+      end
+    end
+
+    it "will not generate the same pdf twice" do
```
K8Sewell, line 67: Can this be pulled since it's an empty test?

FCRodriguez7 (approved):

spec/jobs/generate_pdf_job_spec.rb
martinlovell: Yeah...it does still generate the JSON and make the checksum even with empty authoritative json, but maybe a little unrealistic.  I'll take a look and try to make it a little more realistic by putting in at least something...and verify that the test is working as expected.

martinlovell: I updated the tests in https://github.com/yalelibrary/yul-dc-management/pull/960.  Some comments, but also stubbed the checksum creation, which is better than having that run using the blank authoritative_json.  Thanks for catching that something was off....I didn't fully realize it was running the checksum generation when it didn't need to.

PR #970, opened 2022-03-30
Styled parent object display

FCRodriguez7:
**Story**

The display of individual Parent Objects in Management could be improved. 

Acceptance Criterion
- [x] styling the children's list


edejesusyale (approved):

martinlovell:
I think taking away the overflow auto and max height on the child list is going to make it really long and push down the rest of the properties when there are a lot of children.  (I'm not sure what max-height or max-width auto will do.)

PR #974, opened 2022-04-01
Add checksum check if the file exists already

MaggieZhaoYale:
Co-authored-by: Martin Lovell <martin.lovell@yale.edu>

### DIFF
--- a/app/models/child_object.rb
+++ b/app/models/child_object.rb
@@ -69,7 +69,7 @@ class ChildObject < ApplicationRecord
     #  2. We assume that there is only one access master at a time - BUT we only have one access master pair-tree
     #     across *all* environments (no separation of dev, test, uat, production)
     #     how do we ensure we don't accidentally overwrite something we want to keep?
-    if access_master_exists?
+    if access_master_exists? && access_master_checksum_matches?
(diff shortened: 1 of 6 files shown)


app/models/child_object.rb
```diff
@@ -69,7 +69,7 @@ def copy_to_access_master_pairtree
     #  2. We assume that there is only one access master at a time - BUT we only have one access master pair-tree
     #     across *all* environments (no separation of dev, test, uat, production)
     #     how do we ensure we don't accidentally overwrite something we want to keep?
-    if access_master_exists?
```
mikeapp, line 72: Check that the checksum is not null or blank, or check that it resembles a SHA-1.

app/models/child_object.rb
```diff
@@ -79,8 +79,7 @@ def copy_to_access_master_pairtree
     # Create path to access master if it doesn't exist
     FileUtils.mkdir_p(File.join(image_mount, directory, pairtree_path))
```
mikeapp, line 85: Perhaps check the checksum of the source file.  If it is corrupt, the process will never complete successfully.

mikeapp (approved):

PR #992, opened 2022-04-27
Adds generation and bitstream checks to Preservica sync

K8Sewell:
# Summary
Process for syncing images with Preservica will now check for a new generation or bitstream and download a new binary if either is found.

# Related Ticket
[#2034](https://app.zenhub.com/workspaces/yul-dc-5e50083561915b11fc9e5850/issues/yalelibrary/yul-dc/2034)


app/models/parent_object.rb
```diff
@@ -194,41 +195,32 @@ def upsert_preservica_ingest_child_objects(preservica_ingest_hash)
...
+      co = ChildObject.find_by(preservica_content_object_uri: value[:content_uri])
+      next if co.nil?
+      co.order = value[:order]
+      co.preservica_content_object_uri = value[:content_uri]
+      co.preservica_generation_uri = value[:generation_uri]
+      co.preservica_bitstream_uri = value[:bitstream_uri]
+      co.save
     end
+    # create child records for any new items in preservica
+    create_child_records
```
mikeapp, line 215: Is there an issue with calling `create_child_records`, which calls `array_of_child_hashes_from_preservica` ?  The comment says that the latter method downloads all of the child images from Preservica, but we should only need to download the new or changed (per URIs or checksum mismatch) images during a sync.

mikeapp (changes requested):
The automated review is failing due to repeated setup code in the test related to Preservica.  Could you explore whether this could be refactored?     If it turns out that we need to relax the metric a bit we can.

app/models/parent_object.rb
K8Sewell: There is a scrub of already existing child images [here](https://github.com/yalelibrary/yul-dc-management/pull/992/files#diff-acc421bf5477cf1667d0b61e9ded247f3b42335a2425020920d8c20919ff0384R155) in `validate_child_hashes` that removes any child hashes that are duplicates and this should ensure only new child images are added to the parent.

K8Sewell:
> 

Will work on refactoring the duplicated code and hopefully have this passing soon.

mikeapp (approved):

PR #1046, opened 2022-08-09
Indexes to solr even when metadata is not found

K8Sewell:
# Summary
Even when metadata is not found we still want that information indexed to solr.  This MR adds a step to the setup metadata job to check visibility and index if private.

# Related Ticket
[#2146](https://app.zenhub.com/workspaces/yul-dc-5e50083561915b11fc9e5850/issues/yalelibrary/yul-dc/2146)

### DIFF
--- a/app/jobs/setup_metadata_job.rb
+++ b/app/jobs/setup_metadata_job.rb
@@ -14,6 +14,7 @@ class SetupMetadataJob < ApplicationJob
       parent_object.processing_event("SetupMetadataJob failed to find all images.", "failed")
       return
     end
+    return if metadata_missing(parent_object, current_batch_process, current_batch_connection)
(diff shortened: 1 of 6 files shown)


app/jobs/setup_metadata_job.rb
```diff
@@ -14,6 +14,7 @@ def perform(parent_object, current_batch_process, current_batch_connection = par
       parent_object.processing_event("SetupMetadataJob failed to find all images.", "failed")
       return
     end
+    index_private(parent_object)
```
martinlovell, line 17: Should we bother with the fetch and child jobs if the parent is private, @mikeapp?
If it's made visible or redirected, it will re-run setup_metadata_job and go through the normal process.

mikeapp (approved):

PR #1050, opened 2022-08-16
Revert "Removes double parent in batch process detail for preservica resync"

mikeapp:
Reverts yalelibrary/yul-dc-management#1049


mikeapp:
Ignore, CI failure was transient.

PR #1052, opened 2022-08-24
Adds admin set to batch process table

K8Sewell:
# Summary
Batch processes will now track and display which admin sets are involved in the process.  

# Related Ticket
[#2174]()

# Screenshot
image


app/datatables/batch_process_datatable.rb
```diff
@@ -15,6 +15,7 @@ def view_columns
     # or in aliased_join_table.column_name format
     @view_columns ||= {
       process_id: { source: "BatchProcess.id", cond: :eq, searchable: true, orderable: true },
+      admin_set: { source: "BatchProcess.admin_set", cond: :start_with, searchable: true, orderable: true },
```
mikeapp, line 18: Will `:start_with` work here if we have a list?

mikeapp (approved):

PR #1102, opened 2022-11-30
Adds max request error handling

K8Sewell:
# Summary
Adds error handling for permission requests that exceed the max queue length per permission set.

# Related Ticket
[#2290](https://app.zenhub.com/workspaces/yul-dc-5e50083561915b11fc9e5850/issues/yalelibrary/yul-dc/2290)


app/controllers/api/permission_requests_controller.rb
```diff
@@ -2,22 +2,29 @@
...
   def create
-    request = JSON.parse(request.raw_post)
-    # find parent object
-    parent_object = ParentObject.find_by(oid: request['oid'])
-    # find or create and update permission request user
+    request = params
+    parent_object = ParentObject.find(request['oid'].to_i)
+    pr_user = find_or_create_user(request)
+    permission_set = PermissionSet.find(parent_object.permission_set_id)
+    approver = User.with_role(:approver, permission_set).first
```
mikeapp: Let's leave the approver `null` when the request is created.   We'll use the value to record the user who actually approves the request.

mikeapp (changes requested):

app/controllers/api/permission_requests_controller.rb
K8Sewell: That makes sense.  Just updated it so that user is no longer required to create a permission request.

mikeapp (approved):

PR #1116, opened 2022-12-21
Add delayed job to copy original tiffs to S3

K8Sewell:
# Summary
Adds a delayed job to copy a child object's original tiff to S3

# Related Ticket
[#2319](https://app.zenhub.com/workspaces/yul-dc-5e50083561915b11fc9e5850/issues/gh/yalelibrary/yul-dc/2319)


app/jobs/save_original_to_s3_job.rb
```diff
@@ -0,0 +1,27 @@
...
+    child_object = ChildObject.find(child_object_oid)
+    parent_object = child_object.parent_object
+    if parent_object.visibility == "Private" || parent_object.visibility == "Redirect"
+      raise "Not copying image from #{parent_object.oid}. Parent object must have Public or Yale Community Only visibility."
+    end
+    # check if file already exists on S3
+    raise "Not copying image. Child object #{child_object_oid} already exists on S3." if S3Service.s3_exists_for_download?(remote_download_path(child_object_oid))
+    # copy original to downloads bucket
+    metadata = { 'width': child_object.width.to_s, 'height': child_object.height.to_s }
+    S3Service.upload_image_for_download(Pathname.new(child_object.access_master_path), remote_download_path(child_object_oid), "image/tiff", metadata)
```
MaggieZhaoYale, line 27: I wonder if it is worth to check whether the height or width has valid value. child_object.height or child_object.width can be null. e.g. https://collections.library.yale.edu/management/child_objects/11395410.json, the access_master_path exists and image is in different format and no tiff for this child. Upload will throw an error.

K8Sewell: That's a great point.  I'll had a null check here to be on the safe side.  Thanks!

K8Sewell: I'm not sure if there is a better default than 'unknown' but I could alter this to raise an error instead.  I think I'll bring this up at stand up to see what they would prefer the user to experience.

K8Sewell: The more I thought about it the more an error made sense.  I'll still bring it up in stand up though.

app/jobs/save_original_to_s3_job.rb
```diff
@@ -0,0 +1,29 @@
...
+  def default_priority
+    100
+  end
+
+  def perform(child_object_oid)
+    # check parent visibility is either public or YCO
+    child_object = ChildObject.find(child_object_oid)
+    parent_object = child_object.parent_object
+    if parent_object.visibility == "Private" || parent_object.visibility == "Redirect"
+      raise "Not copying image from #{parent_object.oid}. Parent object must have Public or Yale Community Only visibility."
```
martinlovell: When it's not Public or YCO, it should just log and return from the job instead of raising the exception.   Similarly if the image is already on S3 (but don't log based on the ticket).  If it raises the exception, it will queue the job to retry.

martinlovell (changes requested):

app/jobs/save_original_to_s3_job.rb
K8Sewell: Just updated this with your suggested changes.  Thank you!

martinlovell (approved):

PR #1117, opened 2023-01-03
Bump test coverage

K8Sewell:
# Summary
Coverage dropped below our target of 97% and this MR brings the coverage back up to 96.29%.


config/routes.rb
```diff
@@ -58,7 +58,7 @@
   get 'api/oid/new(/:number)', to: 'oid_minter#generate_oids', as: :new_oid
 
   namespace :api do
-    resources :permission_requests, only: [:create]
+    resources :permission_requests, only: [:create, :update]
```
mikeapp: I think this `:update` route should also be removed?

mikeapp (changes requested):

config/routes.rb
jpengst: Good catch, thanks!

mikeapp (approved):

PR #1119, opened 2023-01-10
Creates API call to save original to S3 job

K8Sewell:
# Summary
Sets up an API that when called will trigger the job to save an original to S3.

# Related Ticket
[#2321](https://app.zenhub.com/workspaces/yul-dc-5e50083561915b11fc9e5850/issues/gh/yalelibrary/yul-dc/2321)

### DIFF
--- a/app/controllers/download_original_controller.rb
+++ b/app/controllers/download_original_controller.rb
@@ -14,7 +14,8 @@ class DownloadOriginalController < ApplicationController
   end
 
   def check_child_visibility(child_object)
-    render(json: { "title": "Child Object is restricted." }, status: 403) && (return false) if child_object.parent_object.visibility == "Private" || child_object.parent_object.visibility == "Redirect"
+    render(json: { "title": "Child Object is restricted." }, status: 403) && (return false) unless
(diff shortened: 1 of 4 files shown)


app/controllers/download_original_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+    rescue ActiveRecord::RecordNotFound
+      render(json: { "title": "Invalid Child OID" }, status: 400) && (return false)
+    end
+    return unless check_child_visibility(child_object)
+    SaveOriginalToS3Job.perform_later(child_object.oid)
+    render(json: { "title": "Child object staged for download." }, status: 200)
+  end
+
+  def check_child_visibility(child_object)
+    render(json: { "title": "Child Object is restricted." }, status: 403) && (return false) if child_object.parent_object.visibility == "Private" || child_object.parent_object.visibility == "Redirect"
```
mikeapp: could we change this to check for statuses that are allowed, so it 'fails safe' if we introduce a new one?

mikeapp (changes requested):

mikeapp (approved):

PR #1138, opened 2023-02-09
Active Permission Set Terms API

jpengst:
## Summary  
`/api/permission_sets/:id/terms` returns the active terms and conditions for the permission set.  
  
## Ticket  
[2376](https://app.zenhub.com/workspaces/yul-dc-5e50083561915b11fc9e5850/issues/gh/yalelibrary/yul-dc/2376)  
  
## Screenshots:  
  
Active Term:  
image
  
Invalid Permission Set:  
image

### DIFF
--- a/app/controllers/permission_sets_controller.rb
+++ b/app/controllers/permission_sets_controller.rb
@@ -77,7 +77,7 @@ class PermissionSetsController < ApplicationController
     end
     # check for terms on set
     if permission_set.permission_set_terms.blank? || permission_set.permission_set_terms.last.inactivated_at.present?
-      render(json: { }, status: 204)
+      render(json: {}, status: 204)
     else
       term = permission_set.permission_set_terms.last
       active_term = term.slice(:id, :title, :body)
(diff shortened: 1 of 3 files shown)


app/controllers/permission_sets_controller.rb
```diff
@@ -68,6 +68,23 @@ def new_term
...
 
+  def terms_api
+    # check for valid permission set
+    begin
+      permission_set = PermissionSet.find(params[:id])
+    rescue ActiveRecord::RecordNotFound
+      render(json: { "title": "Permission Set not found" }, status: 400) && (return false)
+    end
+    # check for terms on set
+    if permission_set.permission_set_terms.blank? || permission_set.permission_set_terms.last.inactivated_at.present?
```
mikeapp: Is the order guaranteed for `last`?  Would it be better to check `permission_set.active_permission_set_terms`?

mikeapp (changes requested):

app/controllers/permission_sets_controller.rb
mikeapp: Nvm, primary key is default.

mikeapp (approved):

PR #1154, opened 2023-03-13
default visibility to private if not valid visibility

jpengst:
## Summary  
If an object does not have a valid visibility, default to "Private"  
  
## Screenshots:  
This is the "Open With Permission" visibility object that was failing on create. Object now gets created successfully with a "Private" visibility:  
image
  
Parent Object Datatable was not counting the child objects because the object was not getting created successfully. Child Object Count is now counted correctly and reflected on the datatable (Object 12479170):  
image

### DIFF
--- a/app/models/parent_object.rb
+++ b/app/models/parent_object.rb
@@ -79,11 +79,8 @@ class ParentObject < ApplicationRecord # rubocop:disable Metrics/ClassLength
     [nil, "Completely digitized", "Partially digitized"]
   end
 
-  # validates :visibility, inclusion: { in: visibilities, allow_nil: true,
-  #                                     message: "%{value} is not a valid value" }
-
   def validate_visibility
-    return true if ['Private', 'Public', 'Redirect', 'Yale Community Only'].include?(self.visibility)
+    return true if ['Private', 'Public', 'Redirect', 'Yale Community Only'].include?(visibility)
 
     self.visibility = 'Private'
   end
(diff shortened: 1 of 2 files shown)


app/models/parent_object.rb
```diff
@@ -78,8 +79,11 @@ def self.extent_of_digitizations
     [nil, "Completely digitized", "Partially digitized"]
   end
 
-  validates :visibility, inclusion: { in: visibilities, allow_nil: true,
-                                      message: "%{value} is not a valid value" }
+  def validate_visibility
+    return true if ['Private', 'Public', 'Redirect', 'Yale Community Only'].include?(visibility)
```
mikeapp: Can we reference `self.visibilities` here instead of repeating the constants?

mikeapp (changes requested):

app/models/parent_object.rb
jpengst: I don't think I can call a self.method in a before_save callback because the object doesn't exist yet so it doesn't know what self is. I can change `def self.visibilities` to `def visibilities` and it works, but it breaks the dropdown in the Create PO form. It looks like `self.visibilities` is only used for that form dropdown so I can change the method to the normal `def visibilities` and add the array directly to the `form.select` dropdown if we want?

jpengst: Actually, it looks like `self.visibilities` is being used in updatatble as well.

mikeapp: I meant access it as a class-level method via `Parent.visibilities`

mikeapp (approved):

PR #1166, opened 2023-03-28
MC get DCS update/create

MaggieZhaoYale:
close #2421 https://github.com/yalelibrary/YUL-DC/issues/2421 #2444 https://github.com/yalelibrary/YUL-DC/issues/2444
(Configure the metadata-test as the metadata_cloud_host
Get Update/Delete/Create steams in metadata_cloud activity:
Screen Shot 2023-03-30 at 12 42 32 PM

### DIFF
--- a/spec/models/dcs_activity_stream_update_spec.rb
+++ b/spec/models/dcs_activity_stream_update_spec.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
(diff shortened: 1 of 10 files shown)


spec/models/dcs_activity_stream_update_spec.rb
K8Sewell, line 1: Can we add some tests here?  Let me know if you would like to pair on them.

mikeapp (approved):

PR #1170, opened 2023-04-05
Allow admin set update in updatable

jpengst:
## Summary  
Users can now update admin set when using Update Parent Objects batch process if they have edit access.  
  
## Ticket  
[2448](https://github.com/orgs/yalelibrary/projects/1/views/1?pane=issue&itemId=24650777)

### DIFF
--- a/app/models/concerns/updatable.rb
+++ b/app/models/concerns/updatable.rb
@@ -76,7 +76,7 @@ module Updatable
       next unless parent_object
       sets << ', ' + parent_object.admin_set.key
       split_sets = sets.split(',').uniq.reject(&:blank?)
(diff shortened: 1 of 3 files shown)


app/models/concerns/updatable.rb
```diff
@@ -76,6 +76,7 @@ def update_parent_objects
       next unless parent_object
       sets << ', ' + parent_object.admin_set.key
       split_sets = sets.split(',').uniq.reject(&:blank?)
+      admin_set = editable_admin_set(row['admin_set'], oid, index) unless row['admin_set'].blank?
```
mikeapp: `editable_admin_set` can return `false`, in which case I think the logic should skip the row?   But it could also be `nil`.

mikeapp (changes requested):

mikeapp (approved):

PR #1173, opened 2023-04-12
Parent show bug

jpengst:
## Summary  
Parent Object showpage is now displaying correctly for minimal metadata objects  
  
## Ticket  
[2450](https://github.com/orgs/yalelibrary/projects/1/views/1?pane=issue&itemId=24688376)  
  
## Screenshot:  
image

### DIFF
--- a/app/assets/stylesheets/parent_objects.scss
+++ b/app/assets/stylesheets/parent_objects.scss
@@ -106,14 +106,6 @@ DEFAULT DESKTOP STYLING
   display: flex;
 }
 
-#parent-show em {
-  grid-area: a;
-}
-
-#parent-show table {
-  grid-area: b;
-}
-
 .show-header {
   font-family: $yale_new_roman, 'Times New Roman', Times, serif !important;
   font-size: 1.5rem;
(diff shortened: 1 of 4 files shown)


public/batch_processes/templates/update_parent_objects.csv
K8Sewell, line 1: I think this file needs to be pulled from this PR.

yarn.lock
K8Sewell, line 1: I think this file also needs to be pulled from this PR.  Looks like it changes the formatting but doesn't change any of the versions so I'm not sure how that applies to this styling PR.

app/assets/stylesheets/parent_objects.scss
```diff
@@ -99,7 +112,6 @@ DEFAULT DESKTOP STYLING
 
 #parent-show table {
```
K8Sewell: You can delete this because you removed the 'grid template' in the containing div so the 'grid area' isn't doing anything anymore.

app/assets/stylesheets/parent_objects.scss
```diff
@@ -88,9 +96,14 @@ DEFAULT DESKTOP STYLING
...
+
 #parent-show {
-  display: grid;
-  grid-template: 'a b' auto;
+  width: 100%;
+  padding-top: 3%;
+  display: flex;
 }
 
 #parent-show em {
```
K8Sewell: This can be deleted as well because the 'grid area' is dependent on the 'grid template' that got removed.

K8Sewell (approved):

PR #1179, opened 2023-04-21
Make uids unique before adding index

martinlovell:
This will make the uids unique if there are duplicates.
(It only works if there are 2 or fewer with the same id, but will eventually work if it's run multiple times.)


martinlovell:
I updated the demo and production manually, so we don't need this PR.

PR #1181, opened 2023-04-21
Aspace update

jpengst:
## Summary  
ASpace objects can now be created through the Create Parent Objects Batch Process  
  
## Ticket  
[2436](https://github.com/orgs/yalelibrary/projects/1/views/1?pane=issue&itemId=23863856)

### DIFF
--- a/app/models/batch_process.rb
+++ b/app/models/batch_process.rb
@@ -207,6 +207,7 @@ class BatchProcess < ApplicationRecord # rubocop:disable Metrics/ClassLength
         oid = row['oid']
         metadata_source = row['source']
         set = row['admin_set']
+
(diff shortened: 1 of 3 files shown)


app/models/batch_process.rb
```diff
@@ -226,6 +227,16 @@ def create_new_parent_csv
           parent_object = ParentObject.find_or_initialize_by(oid: oid)
         end
 
+        aspace_uri = row['aspace_uri']
```
mikeapp, line 230: If the parent is NOT new, and the sheet doesn't contain this column, will this erase the data?   (I don't recall why we allow modification of an existing OID/parent in the `else` block above...)

mikeapp (approved):

PR #1202, opened 2023-05-31
Store extent of full text on child and parent objects

K8Sewell:
# Summary
By storing the extent of full text value in the database it will allow for sorting and searching of the field as expected.

# Related Ticket
[#2489](https://github.com/yalelibrary/YUL-DC/issues/2489)

# Notes
duplication of package.json files from PR https://github.com/yalelibrary/yul-dc-management/pull/1201 that will disappear once merged into main

### DIFF
--- a/app/datatables/parent_object_datatable.rb
+++ b/app/datatables/parent_object_datatable.rb
@@ -38,7 +38,7 @@ class ParentObjectDatatable < ApplicationDatatable
       extent_of_digitization: { source: "ParentObject.extent_of_digitization", cond: :string_eq, searchable: true, options: ["Completely digitized", "Partially digitized"], orderable: true },
       digitization_note: { source: "ParentObject.digitization_note", cond: :like, searchable: true, orderable: true },
       digitization_funding_source: { source: "ParentObject.digitization_funding_source", cond: :like, searchable: true, orderable: true },
-      full_text: { source: "ChildObject.full_text", searchable: true, orderable: true },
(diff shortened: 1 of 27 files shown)


app/datatables/parent_object_datatable.rb
```diff
@@ -81,25 +81,6 @@ def data
   # rubocop:enable Rails/OutputSafety,Metrics/MethodLength
   # rubocop:enable Metrics/AbcSize
 
-  def extent_of_full_text(parent_object)
```
K8Sewell, line 84: this method already exists on the parent object model so it did not need transferred

mikeapp:
Is this using the database fields from the migration?  It might be better to compute the status in the parent and store it rather than iterate as in `extent_of_full_text`.   Perhaps the state could be stored when `update_fulltext_for_children` is called?

K8Sewell:
> Is this using the database fields from the migration? It might be better to compute the status in the parent and store it rather than iterate as in `extent_of_full_text`. Perhaps the state could be stored when `update_fulltext_for_children` is called?

Yes, it's using the database fields.  I thought that would be the best way to ensure the full functionality of the datatable features.  Unfortunately, I'm not sure I understand the two sentences after your question though.  Are you suggesting not adding them to the database and just creating a method in the parent and child models?  One that gets updated during `update_fulltext_for_children`?

mikeapp (approved):

PR #1203, opened 2023-06-01
Requires extent of digitization in csv for creating aspace parent

K8Sewell:
# Summary
When creating a parent object with Aspace as a source the process now requires that the extent of digitization field be included in the csv.

# Related Ticket 
[#2495](https://github.com/yalelibrary/YUL-DC/issues/2495)

### DIFF
--- a/app/models/batch_process.rb
+++ b/app/models/batch_process.rb
@@ -235,10 +235,10 @@ class BatchProcess < ApplicationRecord # rubocop:disable Metrics/ClassLength
         parent_object.holding = row['holding']
         parent_object.item = row['item']
 
-        if row['aspace_uri'].present? && row['extent_of_digitization'].blank?
+        if metadata_source == 'aspace' && row['extent_of_digitization'].blank?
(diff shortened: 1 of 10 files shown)


app/models/batch_process.rb
```diff
@@ -230,15 +230,22 @@ def create_new_parent_csv
...
-        parent_object.aspace_uri = aspace_uri
-        parent_object.bib = bib
-        parent_object.holding = holding
-        parent_object.item = item
+        parent_object.aspace_uri = row['aspace_uri']
+        parent_object.bib = row['bib']
+        parent_object.holding = row['holding']
+        parent_object.item = row['item']
+
+        if row['aspace_uri'].present? && row['extent_of_digitization'].blank?
```
martinlovell: It should check the source to see if the record is using aspace, instead of checking for `row['aspace_uri'].present?`.

check if `metadata_source` or `row['source']` is "aspace"

martinlovell (changes requested):

app/models/batch_process.rb
K8Sewell: Thank you for the catch.  I've just updated the branch with the suggested change.

martinlovell (approved):

PR #1207, opened 2023-06-12
Sierra integration mvp

jpengst:
## Summary  
MVP For new Sierra integration  
  
## Ticket  
[2498](https://github.com/orgs/yalelibrary/projects/1/views/1?filterQuery=&pane=issue&itemId=29263284)

### DIFF
--- a/app/datatables/parent_object_datatable.rb
+++ b/app/datatables/parent_object_datatable.rb
@@ -19,7 +19,7 @@ class ParentObjectDatatable < ApplicationDatatable
     @view_columns ||= {
       oid: { source: "ParentObject.oid", cond: :start_with, searchable: true, orderable: true },
       admin_set: { source: "AdminSet.key", cond: :string_eq, searchable: true, options: @set_keys, orderable: true },
-      authoritative_source: { source: "MetadataSource.metadata_cloud_name", cond: :string_eq, searchable: true, options: ["ladybird", "aspace", "ils"], orderable: true },
+      authoritative_source: { source: "MetadataSource.metadata_cloud_name", cond: :string_eq, searchable: true, options: ["ladybird", "aspace", "ils", "sierra"], orderable: true },
       child_object_count: { source: "ParentObject.child_object_count", orderable: true },
       call_number: { source: "ParentObject.call_number", searchable: true, orderable: true },
       container_grouping: { source: "ParentObject.container_grouping", searchable: true, orderable: true },
--- a/app/datatables/parent_object_datatable.rb
+++ b/app/datatables/parent_object_datatable.rb
@@ -100,10 +100,6 @@ class ParentObjectDatatable < ApplicationDatatable
     "Yes"
   end
 
-  def metadata_sources
-    ["ladybird", "aspace", "ils", "sierra"]
-  end
-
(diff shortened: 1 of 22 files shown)


app/models/parent_object.rb
```diff
@@ -490,10 +496,21 @@ def aspace_json=(a_record)
...
+    self.bib = s_record["bibId"]
+    self.last_sierra_update = DateTime.current
+  end
+
   def ladybird_cloud_url
     "https://#{MetadataSource.metadata_cloud_host}/metadatacloud/api/#{MetadataSource.metadata_cloud_version}/ladybird/oid/#{oid}?include-children=1"
   end
 
+  def sierra_cloud_url
+    "https://metadata-api-test.library.yale.edu/metadatacloud/api/sierra/bib/#{bib}"
```
jpengst: We use `#{MetadataSource.metadata_cloud_host}` for the other source api's but it does not yet exist for sierra. This is hardcoded to `metadata-api-test` for now.

mikeapp: This method should be a copy of `voyager_cloud_url` below, but instead of `/ils` use `/sierra` in the URL string that is returned.

app/models/parent_object.rb
```diff
@@ -490,10 +500,31 @@ def aspace_json=(a_record)
     self.barcode = a_record["orbisBarcode"]
   end
 
+  def sierra_json=(s_record)
+    super(s_record)
+    return s_record if s_record.blank?
+    self.bib = s_record["bibId"]
```
mikeapp, line 508: I think you can also copy these two from the `voyager_json` method:
```
self.item = s_record["itemId"] unless s_record["itemId"].zero?
self.last_id_update = DateTime.current
```

There should not be a holding for Sierra, so we can ignore that.

app/datatables/parent_object_datatable.rb
```diff
@@ -100,6 +100,10 @@ def extent_of_full_text(parent_object)
     "Yes"
   end
 
+  def metadata_sources
+    ["ladybird", "aspace", "ils", "sierra"]
+  end
+
```
mikeapp: This could move to MetadataSource so it is only in one place, and pull from the database.  Something like:
```
def self.all_metadata_cloud_names
  MetadataSource.distinct.pluck(:metadata_cloud_name)
end 
```

app/models/batch_process.rb
```diff
@@ -335,10 +335,10 @@ def update_fulltext_status(offset = 0, limit = -1)
 
   # CHECKS THAT METADATA SOURCE IS VALID - USED BY UPDATE
   def validate_metadata_source(metadata_source, index)
-    if metadata_source == 'aspace' || metadata_source == 'ils' || metadata_source == 'ladybird'
+    if metadata_source == 'aspace' || metadata_source == 'ils' || metadata_source == 'ladybird' || metadata_source == 'sierra'
```
mikeapp: If you implement the method on MetadataSource suggested above, this could be flipped to `MetadataSource.all_metadata_cloud_names.include?(metadata_source)`

mikeapp (changes requested):

mikeapp (approved):

PR #1209, opened 2023-06-15
Adds fields to digital object json

K8Sewell:
# Summary
Adds five more fields to the digital object json and now allows for aspace and ils as source to construct a digital object.

# Related Ticket
[#2506](https://github.com/yalelibrary/YUL-DC/issues/2506)


martinlovell (changes requested):
If should be just "aspace" still, or if you want to add [this issue](https://github.com/yalelibrary/YUL-DC/issues/2507) into this PR, add the check for the feature flag.

martinlovell (approved):

PR #1218, opened 2023-06-29
Updates ArchiveSpace to ArchivesSpace

jpengst:
## Summary  
ArchiveSpace now reads ArchivesSpace


martinlovell (commented):
Everything looks good and continued to work, but if I started with a DB that was already created, the seeds didn't run.  So, the parent object view page is still said "ArchiveSpace."

martinlovell (approved):

PR #1262, opened 2023-09-11
2592 parent object owp views

jpengst:
## Summary  
Visibility/Permission Set controls are now based on Permission Set admin access controls.  
  
 ## Screenshots:  
**_As a non-admin_**:  
  
OwP not an option, Permission Set select disabled:  
image
  
OwP already assigned (visibility and permission set disabled):  
image
  
**_As an Admin:_**  
Can assign OwP as a visibility:  
image
  
Can assign a Permission Set (filtered based on which permission sets the user is an admin of):  
image  
  
Permission Set becomes enabled only after OwP is selected:  
image
  
image

### DIFF
--- a/app/controllers/parent_objects_controller.rb
+++ b/app/controllers/parent_objects_controller.rb
@@ -68,12 +68,13 @@ class ParentObjectsController < ApplicationController
     respond_to do |format|
       parent_object = ParentObject.find(params[:id])
       permission_set = parent_object&.permission_set
+      permission_set_param = OpenWithPermission::PermissionSet.find(parent_object_params[:permission_set_id]) if parent_object_params[:permission_set_id].present?
 
       authorize!(:owp_access, permission_set) if parent_object.visibility == "Open with Permission" && parent_object.visibility != parent_object_params[:visibility]
 
-      authorize!(:owp_access, permission_set) if permission_set.present? && parent_object_params[:permission_set_id].nil?
+      authorize!(:owp_access, permission_set) if permission_set.present? && permission_set_param.nil?
 
-      authorize!(:owp_access, OpenWithPermission::PermissionSet) if permission_set.nil? && parent_object_params[:permission_set_id].present?
+      authorize!(:owp_access, permission_set_param) if permission_set_param.present?
 
       invalidate_admin_set_edit unless valid_admin_set_edit?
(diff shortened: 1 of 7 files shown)


app/controllers/parent_objects_controller.rb
```diff
@@ -59,8 +60,21 @@ def valid_request?
...
   def update
     respond_to do |format|
+      parent_object = ParentObject.find(params[:id])
+      permission_set = parent_object&.permission_set
+
+      authorize!(:owp_access, permission_set) if parent_object.visibility == "Open with Permission" && parent_object.visibility != parent_object_params[:visibility]
+
+      authorize!(:owp_access, permission_set) if permission_set.present? && parent_object_params[:permission_set_id].nil?
+
+      authorize!(:owp_access, OpenWithPermission::PermissionSet) if permission_set.nil? && parent_object_params[:permission_set_id].present?
```
mikeapp: Does this confirm that the user has access to the PermissionSet identified by  `parent_object_params[:permission_set_id]` ?  I think that's the desired case if someone is moving a Parent out of a PermissionSet.

martinlovell: You should remove `permission_set.nil?` from the if too. It should check the specific PermissionSet the user is setting it to regardless of `permission_set.nil?`.

app/controllers/parent_objects_controller.rb
```diff
@@ -59,8 +60,22 @@ def valid_request?
...
+  # rubocop:disable Metrics/PerceivedComplexity
   def update
     respond_to do |format|
+      parent_object = ParentObject.find(params[:id])
+      permission_set = parent_object&.permission_set
+      permission_set_param = OpenWithPermission::PermissionSet.find(parent_object_params[:permission_set_id]) if parent_object_params[:permission_set_id].present?
+
+      authorize!(:owp_access, permission_set) if parent_object.visibility == "Open with Permission" && parent_object.visibility != parent_object_params[:visibility]
+
+      authorize!(:owp_access, permission_set) if permission_set.present? && permission_set_param.nil?
```
mikeapp: I think this check should be performed whenever the permission_set is present and is not equal to the permission_set_param (i.e., whether `nil` or the assignment is to a different permission set).

mikeapp (changes requested):

mikeapp (approved):

PR #1302, opened 2023-12-19
Adds fields to Creat Parent Batch Process

jpengst:
## Summary  
Adds digitization_note, digitization_funding_source, rights_statement, viewing_direction, and display_layout to the Create Parent Objects Batch Process.  
  
## Screenshot:  
image

### DIFF
--- a/app/models/batch_process.rb
+++ b/app/models/batch_process.rb
@@ -237,8 +237,18 @@ class BatchProcess < ApplicationRecord # rubocop:disable Metrics/ClassLength
         parent_object.digitization_note = row['digitization_note']
         parent_object.digitization_funding_source = row['digitization_funding_source']
         parent_object.rights_statement = row['rights_statement']
-        parent_object.viewing_direction = row['viewing_direction']
-        parent_object.display_layout = row['display_layout']
+
+        if valid_viewing_directions.include?(row['viewing_direction'])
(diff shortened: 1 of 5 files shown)


app/models/batch_process.rb
```diff
@@ -272,6 +289,14 @@ def create_new_parent_csv
   # rubocop:enable  Metrics/CyclomaticComplexity
   # rubocop:enable Metrics/BlockLength
 
+  def valid_viewing_directions
```
K8Sewell: I think you can replace this method by using the one that's similar to it on the parent object.  https://github.com/yalelibrary/yul-dc-management/blob/main/app/models/parent_object.rb#L73

app/models/batch_process.rb
```diff
@@ -272,6 +289,14 @@ def create_new_parent_csv
...
+    ['left-to-right', 'right-to-left', 'top-to-bottom', 'bottom-to-top', nil]
+  end
+
+  def valid_display_layouts
```
K8Sewell: I think you can replace this method with something like `parent_object.viewing_hints.include?` https://github.com/yalelibrary/yul-dc-management/blob/main/app/models/parent_object.rb#L78

K8Sewell (approved):

PR #1316, opened 2024-02-01
adds extent_of_digitization to preservica create parent

jpengst:
Preservica objects Extent of Digitization will now be recorded in Create Parent Batch Process


K8Sewell (approved):
Looks great!

PR #1318, opened 2024-02-01
adds terms warning for sets without an active term

jpengst:
## Summary  
Management users will see a Terms warning message whenever a Permission Set does not have any active terms associated with it.  
  
## Screenshot:  
image


K8Sewell (approved):
Looks good

PR #1329, opened 2024-02-21
check for OwP visibility instead of ps_key presence

jpengst:
Batch Update was not triggering OwP validations if PS Key was blank in CSV. We should be checking for OwP visibility and then validate PS Key.  
  
Also adds a check/batch processing event if the PS_Key is blank


K8Sewell (approved):
Looks great!

PR #1331, opened 2024-02-24
Adds unique validation to permission set label

K8Sewell:
# Summary
Ensures that the Label attribute on Permission Sets must be unique.

# Related Ticket
[#2740](https://github.com/yalelibrary/YUL-DC/issues/2740)

# Screenshot

image


martinlovell (approved):

martinlovell:
It might be worth adding an index to the table in the DB.  See the warning here: https://guides.rubyonrails.org/active_record_validations.html#uniqueness

PR #1355, opened 2024-04-10
Store permission request user full name on permission request

jpengst:
## Summary  
Adds `permission_request_user_name` to permission requests in order to save users name they entered in the request form.


K8Sewell (approved):
this looks great

PR #1371, opened 2024-05-07
SPIKE: Adds logging

K8Sewell:


K8Sewell:
Moved debugging to spike #1386

PR #1379, opened 2024-05-13
Adds check for itemPermission

K8Sewell:
# Summary
If an object is submitted for ingest and it is `Open with Permission' then it needs to have a Permission Set Key in the .csv for the batch process.  This PR will add a look at the object's authoritative json to make sure that the object is not an OwP object and will kick back the batch process and request the user submit a .csv with the Permission Set Key included.

Also, included in this PR are some small changes to the source for original images when creating ptiffs.

# Related Ticket
[#2827](https://github.com/yalelibrary/YUL-DC/issues/2827)


K8Sewell:
Split into separate PRs: #1387 , #1385 , & #1386

PR #1348, opened 2024-03-26
Addresses Dependabot Alerts

K8Sewell:
# Summary
Addressess dependabot security alerts.

# Related Ticket
[#2767](https://github.com/yalelibrary/YUL-DC/issues/2767)

PR #1408, opened 2024-06-25
Adds gsub to sanitize for manifests

K8Sewell:
# Summary
During the review for [this PR](https://github.com/yalelibrary/yul-dc-management/pull/1348/files#diff-487c6db6ac726f34f1a91ffbae954366637de8a97d2668cd14420b27256c4569) a change in the behavior of the sanitize method was noticed.  This PR will return the sanitize method to it's desired behavior.

# Related Ticket
[#2818](https://github.com/yalelibrary/YUL-DC/issues/2818)

### DIFF
--- a/app/models/iiif_presentation_v3.rb
+++ b/app/models/iiif_presentation_v3.rb
@@ -144,7 +144,6 @@ class IiifPresentationV3
   # rubocop:enable Metrics/PerceivedComplexity
 
   def sanitize_and_wrap(value)
-    # byebug
     if value.is_a?(Array)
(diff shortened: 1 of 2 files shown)


app/models/iiif_presentation_v3.rb
```diff
@@ -144,10 +144,11 @@ def extract_value(field, hash)
   # rubocop:enable Metrics/PerceivedComplexity
 
   def sanitize_and_wrap(value)
+    # byebug
```
jpengst: comment

jpengst (approved):

app/models/iiif_presentation_v3.rb
K8Sewell: Thank you for catching this!  Removed that comment.

PR #1409, opened 2024-06-26
'Synchronize Child Objects' button access to admin set editors

jpengst:
## Summary  
'Synchronize Child Objects' button on Parent Object showpage can now be accessed by admin set editors.


app/models/ability.rb
```diff
@@ -21,6 +21,7 @@ def initialize(user)
     can :reindex_admin_set, AdminSet, roles: { name: editor_roles, users: { id: user.id } }
     can :crud, ChildObject, parent_object: { admin_set: { roles: { name: editor_roles, users: { id: user.id } } } }
     can :crud, ParentObject, admin_set: { roles: { name: editor_roles, users: { id: user.id } } }
+    can :sync_from_preservica, ParentObject, admin_set: { roles: { name: editor_roles, users: { id: user.id } } }
```
K8Sewell: Since the condition is the same for L24 & L23 you should be able to use the syntax from L26 to allow multiple methods in the same line.

K8Sewell (approved):

PR #1426, opened 2024-08-20
Moves check for OwP auth header

K8Sewell:
# Summary
Moves auth check to before action.

# Related Ticket
[#2877](https://github.com/yalelibrary/YUL-DC/issues/2877)

### DIFF
--- a/app/controllers/api/permission_sets_controller.rb
+++ b/app/controllers/api/permission_sets_controller.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class Api::PermissionSetsController < ApplicationController
-  before_action :check_authorization, except: [:find_or_create_user]
+  before_action :check_authorization
   skip_before_action :authenticate_user!
   skip_before_action :verify_authenticity_token
 
@@ -50,7 +50,6 @@ class Api::PermissionSetsController < ApplicationController
     else
       admin_or_approver_status = "false"
     end
-    response.set_header('Authorization', "Bearer #{ENV['OWP_AUTH_TOKEN']}")
     render(json: { "is_admin_or_approver?": admin_or_approver_status })
   end
   # rubocop:enable Metrics/CyclomaticComplexity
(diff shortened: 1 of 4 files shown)


martinlovell (changes requested):
You should not set the header on the response. 

This would mean that if someone requested without the key, they would get the key in the response...which is supposed to be a secret.

jpengst (approved):

K8Sewell:
> You should not set the header on the response.
> 
> This would mean that if someone requested without the key, they would get the key in the response...which is supposed to be a secret.

Updated this PR to no longer provide auth tokens back to blacklight.

jpengst (approved):

app/controllers/application_controller.rb
```diff
@@ -2,7 +2,7 @@
 
 class ApplicationController < ActionController::Base
   include JwtWebToken
-  before_action :authenticate_user!
+  before_action :authenticate_user!, except: [:check_authorization]
```
martinlovell: You don't need the except for regular methods not tied directly to a route/request. It will only get called on methods triggered because of a route.
(Here and a couple other placed.)

app/controllers/api/permission_sets_controller.rb
```diff
@@ -46,13 +50,14 @@ def check_admin_status
     else
       admin_or_approver_status = "false"
     end
+    response.set_header('Authorization', "Bearer #{ENV['OWP_AUTH_TOKEN']}")
```
martinlovell: One last reponse to remove the header from

martinlovell (changes requested):

app/controllers/api/permission_sets_controller.rb
K8Sewell: removed

app/controllers/application_controller.rb
K8Sewell: Updated this in most recent commit.  Thanks for the catch!

martinlovell (approved):

PR #1469, opened 2024-12-13
Update Integrity Check Error Handling

K8Sewell:
# Summary
Adds ability for user to update the child object's stored checksum if there is a checksum mismatch found during the integrity check job.

# Related Ticket
[#2971](https://github.com/yalelibrary/YUL-DC/issues/2971)

# Screenshot


app/models/child_object.rb
```diff
@@ -96,9 +96,21 @@ def copy_to_access_master_pairtree
   # rubocop:enable  Metrics/MethodLength
   # rubocop:enable  Layout/LineLength
 
-  def access_master_checksum_matches?
-    access_master_checksum = Digest::SHA1.file(access_master_path).to_s
-    checksum == access_master_checksum
+  def checksum_matches?
```
martinlovell, line 99: Since calculating the checksum is pretty expensive, you should calculate them only if they are needed in the if statements.

app/models/child_object.rb
```diff
@@ -96,9 +96,21 @@ def copy_to_access_master_pairtree
...
+    access_sha256_checksum = Digest::SHA256.file(access_master_path).to_s
+    access_sha1_checksum = Digest::SHA1.file(access_master_path).to_s
+    access_md5_checksum = Digest::MD5.file(access_master_path).to_s
+    if checksum.present?
```
martinlovell: Curious on the order.  I was thinking it'd be sha512 (preservica or manually updated by user), checksum (aka sha1 from goobi), sha256 (lb), md5 (lb).

I would put preservica and goobi checksums first since they should always be right unless something is wrong with the file. The ladybird checksums will be wrong in a lot of cases.
Any may be updated from preservica at some point, so the sh512 is the most likely to be right.

app/controllers/child_objects_controller.rb
```diff
@@ -66,6 +66,18 @@ def destroy
     end
   end
 
+  def update_checksum
+    respond_to do |format|
+      if @child_object.update(sha512_checksum: @child_object&.access_sha512_checksum)
```
martinlovell: You could update the filesize here too. It can help with the integrity check and it might be good to have all the sizes.

I wonder if it should clear out the other checksums too, since this should replace them if they are set.

app/controllers/child_objects_controller.rb
```diff
@@ -66,6 +66,18 @@ def destroy
     end
   end
 
+  def update_checksum
```
martinlovell, line 69: Unfortunately, this needs to run on a regular worker, since the management frontend doesn't have access to the file mount.  It's probably good to run it async anyway since it may take a while.
I'd do it with an ActiveJob without a batch job.

martinlovell (changes requested):

martinlovell (approved):

PR #1475, opened 2025-01-15
Segment Large CSV Batch Processes

jpengst:
# Summary
Batch processes, that can potentially have thousands of lines in the CSV, are segmented into batches of 50 to reduce number of objects that have jobs re-run multiple times on the same object.

# Related Ticket
[#2873](https://github.com/yalelibrary/YUL-DC/issues/2873)


app/models/concerns/create_parent_object.rb
```diff
@@ -116,7 +117,9 @@ def create_new_parent_csv
       rescue StandardError => e
         batch_processing_event("Skipping row [#{index + 2}] Unable to save parent: #{e.message}.", "Skipped Row")
       end
+      index + 1 if index + 1 - start_index > BatchProcess::BATCH_LIMIT
```
martinlovell: You need a return on these to break out of the loop. Otherwise it will keep going until it gets to the end of the list.

app/models/concerns/reassociatable.rb
```diff
@@ -45,11 +48,14 @@ def update_child_objects
 
       values_to_update = check_headers(child_headers, row)
       update_child_values(values_to_update, co, row, index)
+      index + 1 if index + 1 - start_index > BatchProcess::BATCH_LIMIT
```
martinlovell: You need the return here too, but with [parents_needing_update, parent_destination_map, index+1]

app/models/concerns/reassociatable.rb
```diff
@@ -7,16 +7,18 @@ module Reassociatable
   BLANK_VALUE = "_blank_"
 
   # triggers the reassociate process
-  def reassociate_child_oids
+  def reassociate_child_oids(start_index = 0)
     return unless batch_action == "reassociate child oids"
-    parents_needing_update, parent_destination_map = update_child_objects
-    update_related_parent_objects(parents_needing_update, parent_destination_map)
+    parents_needing_update, parent_destination_map, denote_continue = update_child_objects(start_index)
+    update_related_parent_objects(parents_needing_update, parent_destination_map, denote_continue)
```
martinlovell: You don't need to pass denote_continue to the update_related_parent_objects().
You can just return it from this method.  update_child_object() keeps track of where it is in the list. update_related_parent_objects() doesn't care about that info.

martinlovell (changes requested):

martinlovell (commented):
Looks good

martinlovell (approved):

PR #1581, opened 2025-08-27
index child captions

jpengst:
## Summary  
Adds caption to child index

### DIFF
--- a/app/models/concerns/solr_indexable.rb
+++ b/app/models/concerns/solr_indexable.rb
@@ -243,20 +243,35 @@ module SolrIndexable
       full_text_array = child_objects.map do |child_object|
         child_object_full_text = S3Service.download_full_text(child_object.remote_ocr_path)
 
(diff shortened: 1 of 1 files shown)


app/models/concerns/solr_indexable.rb
```diff
@@ -243,20 +243,35 @@ def child_object_solr_documents
...
   end
 
-  def child_object_to_solr(child_object, child_object_full_text)
+  def child_object_to_solr_full_text(child_object, child_object_full_text)
     parent_object = child_object.parent_object
     {
       id: child_object.oid,
       parent_ssi: parent_object.oid,
       child_fulltext_tesim: child_object_full_text,
       child_fulltext_wstsim: child_object_full_text,
```
K8Sewell, line 263: Could you please add a _wstsim field for the caption in both methods?

K8Sewell (approved):

PR #1782, opened 2026-08-18
Batch Process Structure Editor

jpengst:
## Summary  
Adds Batch Process for creating and updating parent objects structure ranges.  
  
Validations:  
- Parent exists in the local DB, and the batch's user has update permissions
- Range_name not blank
- Child_oid not blank
- Order_in_range is a non-negative integer
- Range_order is a non-negative integer
- Child exists in the local DB
- Child actually belongs to the parent named in the same row


K8Sewell (approved):
This looks great.  Thank you for all the changes.

